PR #6, opened 2013-01-16
Patch 1

henriquefm:
There was a typo

'henriquefm' - I'm not sure what you did, but what you supplied was a change to every file in the module. What were you trying to change?


mchurchward:
Should be handled by AMOS now.

PR #13, opened 2014-01-23
CONTRIB-4838 Administrators can choose which field is contained in csv download

eberhardt:
Extended the settings page and changed generate_csv()

	modified:   questionnaire.class.php
	modified:   settings.php

### DIFF
--- a/questionnaire.class.php
+++ b/questionnaire.class.php
@@ -2540,33 +2540,23 @@ class questionnaire {
         global $SESSION, $DB;
 
         $output = array();
-        $nbinfocols = 9; // Change this if you want more info columns.
         $stringother = get_string('other', 'questionnaire');
-        $columns = array(
-                get_string('response', 'questionnaire'),
-                get_string('submitted', 'questionnaire'),
-                get_string('institution'),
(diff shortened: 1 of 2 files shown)


mchurchward:
Looking at your pull request, you have a "dirty hack" in the settings.php file. I'm not sure what this is for? The changes work without that. Can you elaborate?
Also, if you choose "submitted" it doesn't actually work. I believe that is due to a typo on line 2803 of questionnaire_class.php: "if (in_array('subbmitted', $options)) array_push($arr, $submitted);" (2 'b's in the array key). Can you confirm?

eberhardt:
Yes, I can confirm. I didn't tested it, but by looking at the code I see the problem. In settings.php it's defined by 'submitted' and I check against 'subbmitted'. That can't work.

The changes at setting.php I defined as dirty hack, because there should be an own string defined in your language packages. I don't know, why I didn't add the string, but I think mostly because I'm not a native english speaker... which means, my translation could be wrong.

It's been a while I requested that pull and therefore I didn't worked with that code a long time. Does the patch apply to newer versions as well?

PR #18, opened 2014-08-19
fix missing $sid param (with $this->survey->id)

nadavkav:


sk-unikent:
Won't ```$this->survey->id``` always be null? This line is wrapped in an ```if (empty($this->survey)) {```

nadavkav:
​Completely right. not sure what I was thinking and why it was working
after that change.
Well... it was long time ago and I am not sure what it was suppose to solve
:-(
As far as I remember, $sid was missing.
Please see if you can dig into that and see if it should be solved a little
bit differently.

PR #21, opened 2014-10-28
trim the comma separated list of email addresses

byzombies:
As far as my testing goes:
email1@example.com, email2@example.com in the current instance the second email address will not receive an email due to the space following the comma.

Just a small issue, took me several minutes to figure out why my second recipient was not getting notifications. Thought I'd save anyone else the hassle.


mchurchward:
Sorry this got overlooked. This will be fixed with https://tracker.moodle.org/browse/CONTRIB-6918

PR #22, opened 2014-11-26
CONTRIB-4891 Questionnaire YES/NO questions results not properly shown

crazyserver:
The same patch can be applied to the other branches

### DIFF
--- a/questiontypes/questiontypes.class.php
+++ b/questiontypes/questiontypes.class.php
@@ -433,9 +433,9 @@ class questionnaire_question {
         }
 
         $sql = 'SELECT choice_id, COUNT(response_id) AS num '.
-               'FROM {questionnaire_'.$this->response_table.'} '.
-               'WHERE question_id= ? '.$ridstr.' AND choice_id != \'\' '.
-               'GROUP BY choice_id';
+                'FROM {questionnaire_'.$this->response_table.'} '.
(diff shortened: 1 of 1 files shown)


timhunt:
The patch introduces unnecessary white-space changes on two lines. You should amend the patch so that those lines are not changed.
Regarding the main part of the change, it is simpler to use sql_empty. Except that, if you look at that method, it recommends that instead you write code like

AND choice_id <> ?

And pass '' in the array of parameters to be put in place of the ?

Wow! this code is pretty bad. It is still concatenating strings to build SQL, and not using placeholders everywhere. E.g. the bit that builds $ridstr should be using $DC->get_in_or_equals(), but I guess that is a separate bug.

remotelearner:
Hi Tim. I implemented the fix in 2.7 and 2.8 using parameter based queries instead. See https://github.com/remotelearner/moodle-mod_questionnaire/commit/8d92d9f54e929c0c8b5ec8c2e2556ed6cf6b4dea#diff-0 for an example.

mike

timhunt:
Ah, great! So this can we close this now?

remotelearner:
When you say "close this", what do you refer to? The tracker issue is resolved (CONTRIB-4891). Is there something else open? (This may be me not being familiar with all of Github's functions).

timhunt:
This pull request. Oh, I can do it.

PR #25, opened 2015-02-19
Contrib 4914 fix drop enum from field

gbateson:
Hi Mike,
here is a pull request for CONTRIB-4914
regards
Gordon


mchurchward:
I'm going to assume that due to the age of this request that it is no longer relevant. If this does still need to be done, please reopen.

PR #26, opened 2015-03-19
Add caching to CSV generation code.

ghost:
A Catalyst client had trouble with browser timeouts while attempting to
get a questionaire report. Execution time for the report was 132s on my
development machine. This patch adds caching of the
questionaire_quest_choice query results in response_select (most of the
savings) and caching of the usernames in generate_csv, reducing execution
time in this case to 23s (an 82% saving).

Signed-off-by: Nigel Cunningham nigelc@catalyst-au.net

### DIFF
--- a/questionnaire.class.php
+++ b/questionnaire.class.php
@@ -1705,6 +1705,8 @@ class questionnaire {
 
     private function response_select($rid, $col = null, $csvexport = false, $choicecodes=0, $choicetext=1) {
         global $DB;
+        $qqc_cache = cache::make_from_params(cache_store::MODE_REQUEST, 'core', 'qqc',
+                array(), array('simplekeys' => true));
 
         $sid = $this->survey->id;
         $values = array();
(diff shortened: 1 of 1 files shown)


timhunt:
If you will pardon the bluntness, that is a pretty nasty hack.

* Caching in static variables is a disaster area when it comes to unit testing. You really should use MUC.
* Only dealing with not having the necessary user data in generate_csv is horrible. We really should JOIN the user table in the original query, and include all the fields we need.

So, this is a short-term fix that massively increases the technical debt in this code. Maybe Mike will be OK with that, but I don't like it.

NigelCunninghamCatalyst:
Thanks for the feedback Tim.

I've heard of MUC but never used it so far. I have however done a reasonable amount of coding in Drupal, where caching in static variables is not so frowned upon. When I get some time, I'll take a look at using MUC instead.

Regarding doing a join in the original query, fair enough. I reduced the ugliness (see the removed-by-this-patch second query of the table), but could have done more to eliminate it entirely :) If I get time to do another iteration of this, I'll look at that too.

NigelCunninghamCatalyst:
Patch updated per feedback - it now uses MUC and the user query has been folded into the previous query instead of using caching.

remotelearner:
Can you create a Moodle tracker item for this? It sounds like we may need more eyes on it, and it should be in the MOODLE_29_STABLE branch going forward.

timhunt:
I just had a quick look and noticed two things:

The cache should be defined in mod/questionnaire/db/caches.php.

I would say that you should not use upper-case characters in a new table alias (https://github.com/NigelCunninghamCatalyst/moodle-mod_questionnaire/commit/a501fa356869dbf5b6fa878fe7c7c1fd2c5e647d#diff-320374203a01c3f0bdc6fa4f7014f46eR2736), even if there are old ones in this part of the code.

NigelCunninghamCatalyst:
Ok. Is there some test framework I can run that checks these things? https://docs.moodle.org/dev/Coding_style is so long that it's hard to check absolutely everything - especially when you work (as I do) on multiple projects with conflicting coding styles!

timhunt:
https://moodle.org/plugins/view/local_codechecker

NigelCunninghamCatalyst:
Thanks!

Nigel

mchurchward:
The CSV generation code was completely rewritten for Moodle 3. I believe that will solve the problem.

PR #27, opened 2015-03-20
Fix XSS vulnerability

gthomas2:
I was able to create a question with javascript in it that alerted document.cookie - replacing 'noclean'=>true with 'noclean'=>false fixed that issue.

### DIFF
--- a/questiontypes/questiontypes.class.php
+++ b/questiontypes/questiontypes.class.php
@@ -928,7 +928,7 @@ class questionnaire_question {
         if ($this->type_id == QUESESSAY) {
             echo html_writer::start_tag('label', array('for' => 'edit-q' . $this->id));
         }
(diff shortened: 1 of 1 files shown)


timhunt:
Security in Moodle is tricky. We have different users with different levels of trust.

This is necessary because teachers may wish to embed usful educational things that require JavaScript.

So, the general rules is that teachers can do XSS but students can't. More accurately, certain capabilities have RISK_XSS included in the risk bitmask, and the default role definitions should only grant those capabilities to teachers and above.

So, this is not a bug, it is by design.

timhunt:
Actually there is a bug, 'mod/questionnaire:editquestions' capability doesn not declare RISK_XSS. I guess all the capabilities should have their risks reviewed.

gthomas2:
Thanks for clarifying this Tim

mchurchward:
Handling in https://tracker.moodle.org/browse/CONTRIB-6919

PR #28, opened 2015-04-15
Remove hard-coded $CFG->prefix

timhunt:
Hi Mike, I spotted some hard-coded $CFG->prefix in the code, so I made this fix.

It will need to be cherry-picked to the other branches. I am happy to do that, if you are happy for this to go in.

### DIFF
--- a/db/upgrade.php
+++ b/db/upgrade.php
@@ -149,7 +149,7 @@ function xmldb_questionnaire_upgrade($oldversion=0) {
 
     if ($oldversion < 2008031904) {
         $sql = "SELECT q.id, q.resp_eligible, q.resp_view, cm.id as cmid
-                FROM {$CFG->prefix}questionnaire q, {course_modules} cm, {modules} m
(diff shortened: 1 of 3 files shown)


remotelearner:
Thanks Tim! Can't believe those were still in there. By all means, please fix up the other branches. Much appreciated.
You have tested your corrections? I'll do so as well.

timhunt:
Oh! testing. Yes, that might be a good idea ;-)
 
I'll test and put on other branches. Do you just want 2.8 & laster, or do you want back-ports as well?

remotelearner:
I think 2.8 and 2.7 will be enough.

timhunt:
Right. Tested and pushed to both 27 and 28 branches.

On the subject of testing, I really wish questoinnaire had Behat tests. If I had the time, I might even make that wish come true, but I don't at the moment. Perhaps when we move to 2.9 I will be able to justify it, or perhaps someone else will beat me to it.

remotelearner:
That would be wonderful of course!

PR #30, opened 2015-05-07
Make alerts more bootstrap compliant and message classes more specific

sharpchi:
The .message class was messing with other .message classes we had on the same page. Specific instances of .message in the CSS have been made specific to #region-main so it won't affect other instances of .message.

I also took the opportunity to change the classes to using the bootstrap alert formats as our users are expecting to see these (bootstrap is used in the clean theme and many others).


remotelearner:
Thanks. Is there a Moodle tracker item open for these improvements? If not can you create one?

mchurchward:
The entire display system has been rewritten to use renderers and templates. If changes are still needed, can you review and resubmit with the new code?

PR #34, opened 2015-05-31
Use new bulk insert support to speed backup.

ghost:
This patch makes use of the support for doing bulk database
record insertions (added in Moodle 2.7) to speed up the
restore of a course. For a questionnaire that had a multichoice
question with options for every suburb in Australia, this patch
was part of a series that reduce the time to restore the course
on my development workstation from 35 minutes to less than 2
minutes.

Signed-off-by: Nigel Cunningham nigelc@catalyst-au.net


remotelearner:
Hi Nigel. Is there a corresponding Moodle tracker item for this issue? If not, could you create one?

PR #41, opened 2015-06-24
Fix deprecated message

keliix06:
Doesn't affect functionality, but when in debug mode it gives a deprecated warning.


mchurchward:
Hi. Thanks for the merge. Could you file a bug in the Moodle tracker in the CONTRIB project with the "Module:Questionnaire" component? Describe what the bug was, how to test it and what the fix does?
Thanks.

PR #42, opened 2015-06-25
M2.9 compatibility fix in mod_form

davosmith:
Switch to standard_intro_elements() instead of add_intro_editor()


davosmith:
Sorry - there's already a pull request for this one.

PR #44, opened 2015-06-30
CONTRIB-5804 - update install.xml to match the current DB structure.

timhunt:


mchurchward:
I'm closing this as I don't think it is the correct fix, as discussed in CONTRIB-5804.

PR #45, opened 2015-07-09
Add scores alongside the given corresponding question in user answer'…

TheHacker66:
…s page


mchurchward:
Could you add an issue to the Moodle tracker for this? Use th CONTRIB project and the component "Module: Questionnaire".
Thanks.

TheHacker66:
Hi, I opened the issue that same day, here:

https://tracker.moodle.org/browse/CONTRIB-5824

mchurchward:
I left a comment in the tracker issue. This is not something we should do.

PR #49, opened 2015-09-15
CONTRIB-5909 Fix syntax to improve cross-db compatibility

danmarsden:
This hasn't been tested as well as I would like but has been deployed to a clients site. Would be good for someone to verify this before you merge it.

### DIFF
--- a/questions.php
+++ b/questions.php
@@ -595,14 +595,9 @@ if ($action == "confirmdelquestion" || $action == "confirmdelquestionparent") {
     // Count responses already saved for that question.
     $countresps = 0;
     if ($qtype != QUESSECTIONTEXT) {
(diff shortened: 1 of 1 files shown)


timhunt:
That is a big step in the right direction, but since $resps is not used in the rest of the function, only $countresps, you should really use $DB->count_records for that.

danmarsden:
good spotting Tim! - fixed and rebased.

I didn't look too closely at the rest of that file... the more I looked, the more I wanted to rewrite.. just never enough time.

timhunt:
Thanks. I will leave it to Mike C to do a final review and merge.

mchurchward:
Thanks guys. This looks fine - and necessary. Still not sure why the old code was giving an error though.

Issue #50, opened 2015-10-05
TinyMCE fails to load in Questionnaire with Essay question type.


Syxton:
TinyMCE does not load properly.  Student's can't add images.

cosmikdust:
is this issue still around? we are running 2.8.2 (Build - 2015031901) of the plugin, wondering if it would be worth upgrading.

mchurchward:
Using 3.2, I just set my preferred editor to TinyMCE, and it appears to work fine. So does Atto.

cosmikdust:
Arghh, any idea why the students can't add images via the tinyMCE module? i mean there is an exact same essay box in Quiz plugin where they can do so (after enabling html editor with file picker in settings).

mchurchward:
There's a lot of work to implement that. The code needs to use file areas for each response. Then backup and restore has to be modified to handle that. Also, the export to CSV function needs to be able to do something with that. Its just not as straightforward as it seems.

mchurchward:
See https://tracker.moodle.org/browse/CONTRIB-6607

PR #51, opened 2015-10-05
Fix for bug #50 file picker breaks

Syxton:


Syxton:
The original commit was not tested thoroughly.  I hadn't fully figured out the Moodle file storage.  This should be good now.

mchurchward:
Hi. Thanks for the merge. Could you file a bug in the Moodle tracker in the CONTRIB project with the "Module:Questionnaire" component? Describe what the bug was, how to test it and what the fix does?
Thanks.

mchurchward:
I believe this is part of https://tracker.moodle.org/browse/CONTRIB-5940

mchurchward:
Due to the code structure changing significantly, this request is no longer valid. If you wish to rework it to the new structure of 3.2 and higher, that would be great.

PR #54, opened 2015-11-11
Replace deprecated methods/classes

sk-unikent:


mchurchward:
Hi. Could you resubmit these as separate pull requests where it makes sense?
I believe all of the "deprecated" functions have been replaced as of 3.1. But you also have a change there that looks like it is part of adding the new search function to questionnaire? That should definitely be a separate issue.
Thanks.

mchurchward:
All of these changes have been added to the code through various issues. Thanks.

PR #56, opened 2015-12-20
CONTRIB-6070 - Add support for DB foreign keys installation and make

AviMoto:
Add support for DB foreign keys installation and make some code standard changes


mchurchward:
Relevant portions have been implemented.

PR #59, opened 2016-01-13
Only enrolled users should show as non responders

Syxton:
Managers should not show up as non responders.  When a questionnaire is anonymous this can cause all managers to receive messages about not responding to a questionnaire.


mchurchward:
This has now been fixed in the latest branches.

PR #65, opened 2016-06-30
CONTRIB - The uploaded image path broken in questions

pramithkm:
The images are not loading after uploaded via TinyMCE. And replacing "or" operand with "||" was not tested properly.


mchurchward:
This is fixed in release 3.1.1.

PR #67, opened 2016-08-26
Add support for date with "/" delimiter

nadavkav:
Our users use date/time format with a "/" delimiter. So we added the above fix to enable proper processing of those kind of date formats.

Please review and see if you can integrate.

Nadav

### DIFF
--- a/classes/response/date.php
+++ b/classes/response/date.php
@@ -84,7 +84,7 @@ class date extends base {
                 // Count identical answers (case insensitive).
                 $this->text = $row->response;
                 if (!empty($this->text)) {
-                    $dateparts = preg_split('/-/', $this->text);
+                    $dateparts = preg_split('/-,\//', $this->text);
                     $this->text = make_timestamp($dateparts[0], $dateparts[1], $dateparts[2]); // Unix timestamp.
                     $textidx = clean_text($this->text);
                     $this->counts[$textidx] = !empty($this->counts[$textidx]) ? ($this->counts[$textidx] + 1) : 1;


mchurchward:
Hi. I don't think your change would work correctly. What is the problem you are encountering that you think this will fix?
The lines you are changing are getting the date parts from the data stored in the database in the "questionnaire_response_date" table. These values are always stored in that table in the form "%Y-%m-%d" (see line 178 in locallib.php).
But, how they are stored in the database does not impact how they are displayed. When they are displayed to users, or when asked to input, the format used is the one defined in the language string "strfdate" in the questionnaire language file.

Please let me know what the specific problem is that you are trying to fix.

mike

nadavkav:
Probably, I was not looking deep enough into this. but our database data was holding dates with "/" and not "-" delimiters, and so I got errors/warnings and undisplayable dates.
After fixing it (the way I did) everything looks fine. 
I guess our data in the db is stored differently because we have a different LOCALE than the one you use.
One that is, probably, holding a different date format, like "%d/%m/%Y". and that is way we have that kind of difference in the actual DB data.

Maybe, to make it more generic, we should figure out the date delimiter from get_string( "strftimedatefullshort", "langconfig") ?

Any maybe it is simply stay as I suggested, is it will work for "-" or "/" delimiters, regardless of locale date format.

What do you think?

mchurchward:
I think we need to figure out why your database is holding them with "/" instead of "-" though. Because from what I can see, questionnaire has "-" hardcoded for data responses.
Assuming you are looking at 3.1, does line 178 of "locallib.php" show "$thisdate = trim(userdate ($thisdate, '%Y-%m-%d', '1', false));" ?
When a new date response is inserted into the database, this is the format that is used. I don't know how your database fields would be anything other than "-".

nadavkav:
I will check it tomorrow, when I am at the campus. And also dive into this more deep. (as I should have)
And get back to you. Thank you for all the quick replys! :smile:

mchurchward:
Hi. Were you able to check into this?

PR #73, opened 2016-11-08
PR CONTRIB-6515 - Fixed bug with CSV export

sarjona:
Our databases are Oracle and we've detected this bug in some cases (with MySQL is working properly). We've tried to find a generic solution (in Oracle, converting TO_CHAR the qrX.response field is also working).

Tested in Moodle 3.1.2 Oracle databases.

This is an improving patch for the PR, which uses sql_order_by_text function instead of sql_compare_text: https://github.com/remotelearner/moodle-mod_questionnaire/pull/68

For some DB, like Oracle, it's necessary to specify "The number of chars to use for this field" (for instance 1000), because CLOB fields don't support direct returning of such fields.

### DIFF
--- a/classes/response/base.php
+++ b/classes/response/base.php
@@ -178,7 +178,11 @@ abstract class base {
         foreach ($extraselectfields as $field => $include) {
             $extraselect .= $extraselect === '' ? '' : ', ';
             if ($include) {
+                if ($field === 'response') {
+                    $extraselect .= $DB->sql_order_by_text($alias . '.' . $field, 1000).' AS '.$field;
+                } else {
                 $extraselect .= $alias . '.' . $field;
+                }
             } else {
                 $default = $field === 'response' ? 'null' : 0;
                 $extraselect .= $default.' AS ' . $field;
--- a/classes/response/multiple.php
+++ b/classes/response/multiple.php
@@ -147,7 +147,7 @@ class multiple extends single {
 
         $userfields = $this->user_fields_sql();
         $extraselect = '';
-        $extraselect .= 'qrm.choice_id, qro.response, 0 AS rank';
+        $extraselect .= 'qrm.choice_id, '.$DB->sql_order_by_text('qro.response', 1000).', 0 AS rank';
         $alias = 'qrm';
 
         return "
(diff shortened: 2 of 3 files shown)


mchurchward:
Thanks for the new request. I'm sorry that this issue is taking so long, but I am struggling to understand the problem still.
The part you are fixing is the select field part of the SQL. You are changing the selection of "qro.response" to "$DB->sql_order_by_text('qro.response', 1000)".
 For discussion, I will simplify the SQL.

In the current select statement, the SQL would look like this:
```
SELECT qro.response
FROM mdl_questionnaire_response_other qro;
```

Does that statement work in Oracle?

timhunt:
Yes, that statement works in Oracle.

The kind of thing that does not work is [WHERE text_column = varchar_column] or [ORDER BY text_column].

mchurchward:
Thanks Tim. Then I don't see how the proposed code change can fix the problem? It is adding the "sql_order_by_text" function to the field select portion of the SQL. And the "qro.response" field is not used in any comparison nor order statement. What am I not understanding?

mchurchward:
Could the problem be caused by different types returned in the "response" column? For the sql, the response column can contain:
- varchar(1),
- longtext,
- null.
Maybe we need to have a "cast" on all "response" returns to ensure that they are all the same?

timhunt:
Right, like with =, you can't mix these types in a UNION or similar.

mchurchward:
Okay. I now understand the need for this. Because we longtext and varchars being returned as UNIONS in the same field, Oracle needs to cast them to the same type.
So, using your latest "$DB->sql_order_by_text('qro.response', 1000)" makes sense. 
Does the "1000" mean that it will only return up to 1000 characters? If so, that could still be a problem for essay question responses.

mchurchward:
@sarjona I added some comments and also modified the "boolean" type response here - https://github.com/remotelearner/moodle-mod_questionnaire/commit/b49abea151506cd9224ee70060ce3cdb46af9f66
Can you let me know if this seems to work fine for your case?

sarjona:
@mchurchward Tested new version of the code with Oracle. We confirm everything is working perfectly!! :-)  Thanks! :-)

PR #74, opened 2016-11-29
Fix restoring course url in Confirmation URL field

jason-platts:
This patch fixes the Confirmation field in advanced settings on restore, so that if you have linked to a url within the course it will be correctly updated.


mchurchward:
Merged into master, MOODLE_32_STABLE and MOODLE_31_STABLE.

PR #85, opened 2017-04-19
QAM-23: Update version numbers to bring all installed plugins to same version

rlcharles:


rlcharles:
@mchurchward The version numbers for 3.1 need to be bumped as there is some people with versions numbers that the one on github

mchurchward:
A couple of problems....
1) I can update the $plugin->version as requested. It won't break anything (except my pride). But, I don't want to change the $plugin->release as requested. The "Build - 2016071101" portion should reflect the date the build is released. That helps make it self-documenting, and differentiates it from the $plugin->version date, which needs to remain the same for the entire major release.
2) Changing the upgrade script as requested will break installations that have already done that step as part of version 2016020204. Anyone upgrading from the previous version 2015051102 to your requested version 2016071101 will go through the 2016020204 step as well, so it should remain as is.

rlcharles:
I've removed changes to the upgrade steps.

mchurchward:
I'm going to do a new release with some new function. I'll update with your required version then. Can I confirm that the version you need is 2016071101?

PR #86, opened 2017-04-28
Added composer.json

michaelmeneses:
Allows to add the plugin as a dependency of Moodle

Example
https://github.com/danmarsden/moodle-mod_attendance/blob/master/composer.json


mchurchward:
Did this with an earlier pull request. Thanks.

PR #94, opened 2017-07-10
CONTRIB-6973 mod_questionnaire: Check group permissions in report

barryo-learningpool:


mchurchward:
The patch you suggest will prevent anyone who has the ability to view all groups (has capability accessallgroups) from accessing them if they are not members of all groups, so I don't think that will work either. I believe this whole chunk of code needs to be reworked to get rid of technical debt.

mchurchward:
Closing this due to the amount of time passed.

Issue #100, opened 2017-09-11
Accesibility thread


Tazoeur:
Hi,

A client recently asked us to improve the accessibility of their moodle. They did an audit of their platform (I think its was following the RGAA 2.2 or 3.0 standart) and we had to work on the problems they puted in light.

Since this moodle platform do use your plugin (great work by the way :D ), we though it might be a good idea to share our work with you and the community.

So here is some description of the problems we've faced and I will provide the pull requests in a near future.

## 1. Likert scale - make relation between grid header title and each response

The idea was to link the header with every `<input type="radio">` of the same column.
To do so, we added an id for every header (named q<question_number>_resp<response_number>) and we used the HTML property `aria-labelledby` to do the link.

We also added a `role="presentation"` to the table.

So the desired HTML looks like this : 

```
<table style="width:99.9%" role="presentation">
    <tbody>
      <tr>
        <td id="q1_resp1">religiously agree</td>
        <td id="q1_resp2">agree</td>
        <td id="q1_resp3">meh</td>
        <td id="q1_resp4">disagree</td>
        <td id="q1_resp5">outrageously disagree</td>
      </tr>
      <tr class="raterow" role="radiogroup group" aria-labelledby="q1">
        <td id="q1" >Sharks are stronger than gorillas</td>
        <td style="text-align:center" class="c1 raterow"><input aria-labelledby="q1_resp1" name="q1977_4965" value="0" id="q1977_4965_0" type="radio"></td>
        <td style="text-align:center" class="c1 raterow"><input aria-labelledby="q1_resp2" name="q1977_4965" value="1" id="q1977_4965_1" type="radio"></td>
        <td style="text-align:center" class="c0 raterow"><input aria-labelledby="q1_resp3" name="q1977_4965" value="2" id="q1977_4965_2" type="radio"></td>
        <td style="text-align:center" class="c1 raterow"><input aria-labelledby="q1_resp4" name="q1977_4965" value="3" id="q1977_4965_3" type="radio"></td>
        <td style="text-align:center" class="c0 raterow"><input aria-labelledby="q1_resp5" name="q1977_4965" value="4" id="q1977_4965_4" type="radio"></td>
      </tr>
...
</table>
```
Here is an ugly output of a likert scale question : 

<table style="width:99.9%" role="presentation">
    <tbody>
      <tr>
       <td></td>
        <td id="q1_resp1">religiously agree</td>
        <td id="q1_resp2">agree</td>
[...]

PR #101, opened 2017-09-12
Issue #100 - likert scale : relation bt grid header and radio buttons

Tazoeur:
* adding role="presentation"
* get an unique key for the differents propositions inside a likert scale question (basically a simple enumeration)
* link headers and radio buttons with this unique key

### DIFF
--- a/classes/output/renderer.php
+++ b/classes/output/renderer.php
@@ -178,6 +178,17 @@ class renderer extends \plugin_renderer_base {
         // If the question has a template, then render it from the 'qformelement' context. If no template, then 'qformelement'
         // already contains HTML.
         if (($template = $question->question_template())) {
+	     // here we add some informations to the row
+            $pagetags->qformelement->qnum = $pagetags->qnum;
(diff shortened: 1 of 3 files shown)


mchurchward:
Hi. Can you describe what this fixes and/or improves?

Tazoeur:
If I remember well, this question is like a 2 entries table, with values ('very good', 'good', 'meh', 'bad', 'very bad') on the columns et the questions on the rows.

> The idea was to link the header with every <input type="radio"> of the same column.
> To do so, we added an id for every header (named q<question_number>_resp<response_number>) and we used the HTML property aria-labelledby to do the link.

This solution seems to fix an issue with screenreaders not beeing able to read the description (the tilte of the column) when standing on the radio button.

mchurchward:
Hi. Again, sorry for the delay. If this is an accessibility issue, I am very interested in implementing it. But I am unclear how to verify it. Can you provide some details as to how I can test that it works?

Tazoeur:
Hi

wow, didn't expect to see that kind of content in my mailbox since awhile :)

I do not know how this can be tested, I had to implement it for an accessibility audit but I am do not know how they did perform these kind of tests :s

It seems to really help people with a screen reader to have, for each question, an enumeration of the possible answers so I assume that some kind of checker that the attribute `aria-labelledby` is present for each radio button and that the given id can be found somewhere in the DOM.

Good luck :)

PR #102, opened 2017-09-12
Issue #100 - Required fields

Tazoeur:
* create string 'requiredfield'
* put an 'alt' attribute inside the '' tags
* add a legend (new string) to inform that the required fields are annoted with an asterisk
* change an image_url (deprecated) to a pix_url


Tazoeur:
I don't know why I did replace image_url by pix_url...
I've been recently spending so much time updating moodle plugins to change calls from pix_url to pix_icon and image_url...

The idea is to explain to users what is the meaning of the asterisk.
If the asterisk is not displayed in the survey display functions, you are right and there is no need to add the legend.

PR #105, opened 2017-10-04
Added support for block_myoverview in Moodle 3.3+

ndunand:
cf. https://docs.moodle.org/dev/Calendar_API#Action_events

According to Moodle 3.3 documentation, the new Course Overview block displays notifications in My Page when a teacher sets an 'Expect completed on' date in the activity completion settings. This works will core modules and should work consistently with 3rd-party modules as well.

see https://docs.moodle.org/33/en/Course_overview 

The proposed changes would implement this in mod_questionnaire

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -138,8 +138,10 @@ function questionnaire_add_instance($questionnaire) {
 
     questionnaire_set_events($questionnaire);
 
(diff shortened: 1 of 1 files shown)


mchurchward:
Cherry-picked into 3.3. Will merge into master and 3.4.

PR #131, opened 2018-04-02
Better visualization of the results of checkbox questions, to ease interpretation

ba-al:
In effect of these changes the visualizations at the result page use the number of participants instead of the number of responses per option.

### DIFF
--- a/classes/response/display_support.php
+++ b/classes/response/display_support.php
@@ -67,14 +67,18 @@ class display_support {
                     arsort($weights);
                     break;
             }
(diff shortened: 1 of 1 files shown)


mchurchward:
Can you clarify what this does? Show an example of before and after?

mchurchward:
Ah. Wait. I just noticed the reference to issue 135, which explains it a bit better.

mchurchward:
Can you look at my last comment on https://github.com/PoetOS/moodle-mod_questionnaire/issues/135 for a different implementation?

PR #140, opened 2018-06-26
Show incomplete solutions in CSV

ba-al:
Show incomplete solutions in CSV too (with an additional column "complete" with "y" or "n") so jumping-off points can be discovered. This is coupled with a tick on the export page.


mchurchward:
This has been dealt with in the latest commit and forthcoming release.

PR #146, opened 2018-08-13
added print option to the summary page

BlackNetworkBit:
implemented print option for the summary page.


mchurchward:
There is new work coming soon that adds options to print the various report pages in various formats. I will close this. If after that release, this has still has merit, please resubmit it.

PR #152, opened 2018-09-25
Enrolled users only

Syxton:
This setting means that non enrolled but capable users will either be allowed to take the questionnaire or will not be allowed to take the questionnaire.  It also keeps these users on/off the non-responders list.


Syxton:
Fixes CONTRIB-5429

Issue #162, opened 2018-11-21
Filter responses by group don't return any result


reskit:
Hi,

A questionnaire module configured with visible or separated group mode don't return any result on report page when filtering by group.

I'll provide a patch to solve this problem.

mchurchward:
Thanks for the report. This was fixed in the latest release, 3.5.1.

PR #164, opened 2018-11-21
Fix SQL query to work with groups, use the correct groupid field.

reskit:
Related to issue #162 

There's an error in a couple of SQL querys using id field from groups_members to match the current group when groupid field is the correct one.


mchurchward:
This was fixed in the 3.5.1 release.

mchurchward:
I see that now. Odd, I was sure that had been dealt with. I'll merge your change.

PR #165, opened 2018-11-26
CONTRIB-7412 improve response values when exporting "other" responses.

paulholden:


mchurchward:
Thanks. I will look at this.

mchurchward:
I have cherry-picked these into the MOODLE_35_STABLE branch. This is a good improvement. I think something similar needs to be done for checkboxes, but they are handled differently.

PR #166, opened 2018-11-26
Backport Privacy API support to 3.3 branch

paulholden:
The first commit (5e77dd136aab6c35411ef4552f637030ad8512dc) just contains two backported changes from the 3.5 branch that implement the Privacy API (I've preserved @mchurchward as the author)

The second commit (2c778013001360a0f67f084337cbf80cb3176223) is to fix compatibility with Moodle 3.3 which is the last version to support PHP 5.6 and as such does not have support for most type hinting that is available in later versions of PHP. This was apparently attempted in 89199871fc7c17ced07cada11efa874c0131d1b3 but was not done successfully

### DIFF
--- a/classes/privacy/provider.php
+++ b/classes/privacy/provider.php
@@ -42,7 +42,7 @@ class provider implements
      * @param   collection $items The collection to add metadata to.
      * @return  collection  The array of metadata
      */
-    public static function _get_metadata(\core_privacy\local\metadata\collection $collection): \core_privacy\local\metadata\collection {
+    public static function _get_metadata(\core_privacy\local\metadata\collection $collection) {
 
         // Add all of the relevant tables and fields to the collection.
         $collection->add_database_table('questionnaire_attempts', [
@@ -113,7 +113,7 @@ class provider implements
      * @param   int $userid The user to search.
      * @return  contextlist   $contextlist  The list of contexts used in this plugin.
      */
-    public static function _get_contexts_for_userid(int $userid): \core_privacy\local\request\contextlist {
+    public static function _get_contexts_for_userid($userid) {
         $contextlist = new \core_privacy\local\request\contextlist();
 
         $sql = "SELECT c.id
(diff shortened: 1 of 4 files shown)


mchurchward:
Hi Paul.
I don't really want to backport to the 3.3 branch. I have made the 3.5 branch work with 3.4 and 3.3. Your fixes are necessary to make this actually work - thanks!
My strategy is to introduce new features into the latest branch release only. When the branches can be compatible with earlier branches, I note that in the release notes and make it so.
Non-latest branches get bug fixes only.
I have figured out how to get Travis to test with multiple branches of Moodle and different PHP versions, so that uncovered the error you identified.
I recreated your changes in the latest 3.5 commit. If you want to create a new pull request against the 3.5 branch for those fixes, I'll undo mine and accept yours.

paulholden:
Hi @mchurchward,

> I have made the 3.5 branch work with 3.4 and 3.3. Your fixes are necessary to make this actually work - thanks!

Your 3.5 branch **doesn't work** in most older Moodle versions since 485dc9a8b6e7c8550d4b24a47d7108b26e4c4ec4 as you are implementing an interface (_\core_privacy\local\request\core_userlist_provider_) that doesn't necessarily exist in those versions, and will create a PHP fatal error (all of 3.3, <3.4.6, <3.5.3)

The [release notes](https://docs.moodle.org/dev/Moodle_3.6_release_notes#Privacy_API_update) indicate what versions this will work in

This PR would at least allow users to comply with GDPR while on the 3.3 branch, where your 3.5 branch can't be backported to. Users using 3.4.6 and 3.5.3 would be able to use your 3.5 branch as now

mchurchward:
Hi Paul -
Yes, I understand that, and thanks for that fix. But the privacy API is not in the 3.3 branch, whereas your pull requests wants to put it into the 3.3 branch. The change also contains changes to the notification system (I believe), which should not go back to 3.3.
However, your request should go into both the 3.5 and 3.4 branches.
In any case, I copied your changes to both the 3.5 and 3.4 branches, updated the Travis CI files so they test them with PHP 5.6, and released them.
3.5.2 release - https://github.com/PoetOS/moodle-mod_questionnaire/tree/3.5.2
3.4.3 release - https://github.com/PoetOS/moodle-mod_questionnaire/tree/3.4.3

PR #175, opened 2018-12-18
CONTRIB-7575 : Fixing unanswered dropdown display

ak4t0sh:


mchurchward:
I made some changes yesterday that I committed today. They set the dropdown to show no selection, which matches what the rest of the question types do.

I think your change might be better, but then I think we should make the same sort of change to all questions when they haven't been answered. Thoughts?

PR #177, opened 2019-01-22
Update styles.css

eglescout:
This UI change make checkboxes and radio buttons much easier to click or tap by making the whole table cell clickable instead of just the radio button or checkbox.


mchurchward:
Hi. There was a change made in 3.5.1 (see the release notes), that made the entire radio button cell and checkbox cell clickable. Does this change do something more? If so, can you describe the before and after?

Issue #191, opened 2019-04-19
Tab in "!other" answer is not replaced. 


papillon326:
add "preg_replace("/[\r\n\t]/", ' ', $responsetxt1)"

mchurchward:
Can you provide detail on how to replicate and see the problem this fixes? Thanks.

papillon326:
1. Add new questionnaire to course.
2. Create "checkbox" (or "radio") type question including "!other=" selection.
3. Copy text message with tab from text editor.
4. Answer the question and paste text message into "!other=" text field.  
5. Export CSV.

mchurchward:
Fixed in latest MOODLE_37_STABLE

PR #192, opened 2019-04-20
fix issue #191

papillon326:


mchurchward:
Pulled into latest MOODLE_37_STABLE

PR #193, opened 2019-04-22
contrib-189: Added in check to ensure that $questionobj has data befo…

2shediac:
…re calling get->choices.


mchurchward:
Please test this against release 3.5.4 which supports Moodle 3.3 through 3.6. Also, please provide steps to replicate the problem before the fix.

mchurchward:
Closing due to age.

PR #217, opened 2019-09-03
Corrected Non-respondents list showing every user on system level

doiphode:
The update will show only students enrolled on the course for Non-respondents, rather than all users with roles on system level that have the capability mod/questionnaire:submit set to Allow.


mchurchward:
I'm not sure this is completely correct. Is the "get_users_by_capability" still required in this case?

mchurchward:
This has been completed with https://github.com/PoetOS/moodle-mod_questionnaire/pull/436

PR #221, opened 2019-09-24
Anonymous questionnaire breaks anonimity in individual responses.

scippie75:
When a questionnaire is set to anonymous the name of the user who responded is still shown in the overview of the individual responses.
image001


mchurchward:
Thank you for this. I redid the merge for the MOODLE_37_STABLE branch instead. Your commit is now in that branch.

PR #222, opened 2019-09-24
Rename README.txt to README.md

rahulkhatri19:
For better reading md file will be better.


mchurchward:
Thinking this probably needs its content updated to, to exclude specific version upgrade into, and point to the "CHANGES.txt" for that.

PR #250, opened 2019-11-13
Add fields for action event to calendar events

joostvanderborg:
Action events need to have these two fields set to be shown in for example the timeline-block.  Would this change be possible?

$event->timesort (for sorting the event in a chronological list)
$event->type = CALENDAR_EVENT_TYPE_ACTION to define it as an action event.

See https://docs.moodle.org/dev/Calendar_API#Action_events for details.

I've added the timesort where it seemed logical, but this might not be the case, and could be tested further.


mchurchward:
Pulled into MOODLE_39_STABLE

PR #251, opened 2019-11-20
Style the 'Answer the questions...' link to show as a button.

lucaboesch:


mchurchward:
Merged. But I wonder if the "Your responses" and "View all responses" links should be restyled too?

adpe:
Hi @mchurchward 

Yes they should be styled too and maybe align them `inline`?

lucaboesch:
Adrian, I concluded that it is too hard with the current structure to add an opening
`<div class="btn-group">`
and a closing
`</div>`
in the current coding structure where all gets inserted div-wise.
But the buttons are nicely styled and do stack on the top of each other with a margin.

grafik

See the pull request here: https://github.com/PoetOS/moodle-mod_questionnaire/pull/254

PR #258, opened 2019-12-04
Remove html strong element

adpe:
It is useless and not needed (rework slightly #251)


mchurchward:
Cherry picked into MOODLE_37_STABLE

PR #155, opened 2018-10-26
Add questionnairename field option in export CSV

adpe:

PR #259, opened 2019-12-04
Add new requested and pending downloadoptions

adpe:
As you mentioned here https://github.com/PoetOS/moodle-mod_questionnaire/pull/155#issuecomment-551992530 I've rework these issues and I think now we have the cleaner solution as before.

We're always improving and learning.

Cheers,
Adrian


adpe:
Hi @mchurchward 

I think you dismissed this PR?

Thanks, Adrian

mchurchward:
Pulled into MOODLE_39_STABLE

PR #257, opened 2019-12-04
Align items and remove useless css

adpe:
More information in the commit message

PR #262, opened 2019-12-09
Align answer options (radio buttons)

lucaboesch:
Let's align the radio buttons which otherwise would be cut.

alignansweroptions

What the css does is indenting all input of type radio except if they are preceeded by option (that's when they are horizontally one option after another).

### DIFF
--- a/styles.css
+++ b/styles.css
@@ -262,6 +262,14 @@ td.selected {
     overflow: auto;
 }
 
+.qn-answer input[type=radio] {
+    margin-left: 0.5em;
+}
+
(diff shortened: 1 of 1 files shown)


adpe:
Hi Luca

That's a duplicate of my PR done last week (see https://github.com/PoetOS/moodle-mod_questionnaire/pull/257)

Greets,
Adrian

mchurchward:
Should the same style change be applied to checkboxes?

mchurchward:
Cherry-picked into latest MOODLE_37_STABLE.

Issue #263, opened 2019-12-10
Download responses data as text (cvs) format - buffer not empty


mcardle:
In MOODLE_36_STABLE branch (and maybe others too) when downloading responses as CSV we sometimes get an exception error in the download_as_dataformat() function since it checks for the buffer length and it is not empty.

We have made a fix by adding

```
// Sometimes the buffer is not empty and this will throw an exception in the download_as_dataformat() function.
ob_end_clean();
```
on line 490 in report.php

I will make a pull request for this issue in a bit.

mcardle:
I should mention, that I we are currently using Moodle 3.6.6+ (Build: 20191016) on that site along with theme_clean (not that I think it matter)

mcardle:
Pull request created

https://github.com/PoetOS/moodle-mod_questionnaire/pull/264

mchurchward:
Left a comment on the pull request. The buffer issue should never happen. Can you provide a test case?

PR #266, opened 2019-12-10
#263 - master

mcardle:
Fixed exception "Output can not be buffered before calling download_as_dataformat" thrown in report.php


mchurchward:
There really should never be a time when the output is not empty, unless some other error had occurred. Can you provide an example of when this has happened?

mchurchward:
I think this is unnecessary, but if you can create a test case where this happen, please resubmit.

PR #277, opened 2020-02-25
Backup question choices ordered by their id

MartinGauk:
Otherwise the choices might be restored in random order.


mchurchward:
Cherry picked into MOODLE_38_STABLE

PR #279, opened 2020-03-09
Add a progress bar

clransom:
This adds a simple progress bar at the top of multi-page questionnaires to make it easier to see how many questions there are left to complete.

Features:
- progress percentage is based on number of questions in the whole questionnaire compared to the number of questions on the pages before the current page.
- progress bar only appears on questionnaires that have more than one page
- there is a setting to turn the progress bar on or off per questionnaire (default off)


mchurchward:
Pulled into the MOODLE_39_STABLE branch.

PR #284, opened 2020-04-20
Adding support for multilanguage in choice options.

lucaboesch:
Covered by Behat test extension.
Note: the Moodle App questionnaire does render in the correct language.

Best,
Luca

### DIFF
--- a/classes/question/drop.php
+++ b/classes/question/drop.php
@@ -96,7 +96,7 @@ class drop extends question {
             }
             $option = new \stdClass();
             $option->value = $key;
-            $option->label = $choice->content;
+            $option->label = format_string($choice->content);
             if (isset($response->answers[$this->id][$key])) {
                 $option->selected = true;
             }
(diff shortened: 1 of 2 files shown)


mchurchward:
Hi @lucaboesch . I am interested in this change, but the Behat test still confuses me. I think, what we want, is to see a label displayed in another language, without impacting the choice made. Correct? Can you create a test that reflects that?
Or, for now, create a set of test steps I can try to see this work correctly?

lucaboesch:
Yes, that's exactly it, Mike.
The steps to reproduce the results are unmodified are not in Behat, they do not need to be there.
Behat just makes sure the filter is applied when enabled.
If you wanted to test this, go like that:

1. enable the spanish language pack
2. enable the multilanguage filter
3. add a questionnaire with a pulldown question having the values 1=One,2=Two,3=<span lang="en" class="multilang">Three</span><span lang="es" class="multilang">Tres</span>,4=Four
4. As student1 go and fill out the questionnaire in english choosing Three
5. As student2 go and fill out the questionnaire in english choosing Tres
6. Go and view the results and verify the third option got chosen two times.

Best,
Luca

mchurchward:
Cherry-picked into MOODLE_39_STABLE.

PR #133, opened 2018-04-19
CONTRIB-7264: Add option to exclude inactive users' responses

tonyjbutler:
This commit adds the option to exclude inactive users' responses from questionnaire reports, in line with an increasing number of core Moodle activity modules.

PR #297, opened 2020-06-26
CONTRIB-7264: Add option to exclude inactive users' responses

tonyjbutler:
This commit adds the option to exclude inactive users' responses from questionnaire reports.

New pull request against MOODLE_39_STABLE as per Mike's request at https://github.com/PoetOS/moodle-mod_questionnaire/pull/133#issuecomment-649733335.


mchurchward:
A couple of things...

1. The PR is still conflicting with the MOODLE_39_STABLE branch.
2. I'm uncomfortable with the extra SQL generated for the downloads. The "in_or_equal" bit could be quite large, when there are a large number of responders, and may not work at all.
3. I'm thinking the option should be more general. So when looking at all responses on the page, it should exclude inactive users. Does it?
4. The epic that was referred to in CONTRIB-7264 never seemed to have been completed.

Issue #153, opened 2018-10-08
Add new field to downloadoptions (csv export)


adpe:
Hello @mchurchward 

Would it be possible to add a new field to the downloadoptions? Maybe it would be great to add all available fields and let the user select which ones he wants.

At the moment the needed field is not available for us. We helped us out and added this field, but it's not in the official repository.

mchurchward:
Is this the idnumber request?
I am going through all the pull requests provided over the next weeks, and I will be including them in the next release if they are complete.

adpe:
No it's a different one. 

What I was thinking is, if you want add each downloadoption "hardcoded" to the `configmultiselect` or if there would be another solution with a textarea, where the fields wanted can be added/removed dynamically.

I'll post my solution based on issue #120 , and will be glad to hear from you.

adpe:
Look @ pull request https://github.com/PoetOS/moodle-mod_questionnaire/pull/155

adpe:
Thanks @mchurchward 

Therefore I'm closing this issue!

Issue #120, opened 2017-12-22
Add idnumber field option in export CSV 


papillon326:

adpe:
Hi @papillon326 

It should be now possible for you to use this id number. There are now two new exportoptions!

@mchurchward: I think this issue can be closed?

HTH, Adrian

mchurchward:
I had pulled this change in, but I now realize it was redundant. Adding the database 'id' field for the user has been part of the export for a long time. It is listed as the "ID" option.
I am going to back this change out.
Or perhaps, this was meant to be the 'idnumber' field of the user?

papillon326:
Yes, 'idnumber ' mean 'student idnumber' not 'course module id'.

PR #302, opened 2020-07-14
Bugfix: set course questionnairename if survey is public

adpe:
Hi @mchurchward 

Here was a bug, that when the survey was public, always the survey's modulename was set. But instead we want to set the questionnaire name within each course as modulename, if we're on survey public mode.

HTH,
Adrian

### DIFF
--- a/questionnaire.class.php
+++ b/questionnaire.class.php
@@ -2922,9 +2922,10 @@ class questionnaire {
         if (!$this->survey_is_public()) {
             $courseid = $this->course->id;
             $coursename = $this->course->fullname;
+            $modulename = $this->name;
         } else {
             // For a public questionnaire, look for the course that used it.
-            $sql = 'SELECT q.id, q.course, c.fullname ' .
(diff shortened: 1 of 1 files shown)


mchurchward:
I had to remove your pull request that covered #120 and #153 because the user id was already part of the export as the "ID" field. Since this was combined with the "modulename", that one was removed as well. Since this PR affects the "modulename", you should rework this PR with the addition of the "modulename" option for export.

adpe:
Hi @mchurchward 

Yes it's just like you said. But it was working before. We are using this additional row (downloadoption) in our `fork` for more than one year now... So there must be upstream changes and in due of that, I reworked this existing solution.

So I'll do the rebase with the new working solution and that should be okay? Or must be there a setting, so that you can select if you want add overall the public parent questionnaire modulename instead of the child instances modulenames?

Greets,
Adrian

adpe:
Hi @mchurchward 

As you deleted the previos merged stuff, I added a completely new PR #307 for that and will close this one here, as it will be included in the other.

Greets,
Adrian

PR #303, opened 2020-07-14
Add new downloadoption for participantscount

adpe:
Hi @mchurchward 

We've introduced a new downloadoption, so that you can retrieve the amount of enrolled students from the questionnaires course. It's cool to know how many students were enrolled, and how many responses you get.

Check it out,
Adrian


adpe:
Hi @mchurchward 

Yes it would be the same value, as in our case we need to know how many times the public questionnaire was answered in the particular course.  Maybe that's not the way the plugin should work and we should use `private` ones. What do you think? The average is for us not so important, as we can calculate this afterwards, but I see your point. That when you sum all rows you will get too much responses.

Right now, I check in a script, if the given course was already count, and if yes, then we skip this field on the next responses column. How could we achieve that within a export? Do you have any idea?

Greets,
Adrian

adpe:
This PR must be merged after #307. As it's depending on the additionalchoices array which was introduced in the previous commit. So please first check out #307 and then this PR.

HTH,
Adrian

PR #304, opened 2020-07-14
Bugfix: add complete row if including incomplete responses

adpe:
Fixes regression of commit https://github.com/PoetOS/moodle-mod_questionnaire/commit/9d4a078379ddd41c56c7d83843f066a979f88e26

Why did you delete this complete field? As now it will always be false (0) and therefore this row is never included in the export.

HTH,
Adrian

### DIFF
--- a/templates/extrafields.mustache
+++ b/templates/extrafields.mustache
@@ -25,4 +25,5 @@
     }}
 <input type="checkbox" name="choicecodes" id="choicecodes" checked="checked" value="1" /> <label for="choicecodes">{{#str}}includechoicecodes, questionnaire{{/str}}</label><br />
 <input type="checkbox" name="choicetext" id="choicetext" checked="checked" value="1" /> <label for="choicetext">{{#str}}includechoicetext, questionnaire{{/str}}</label><br />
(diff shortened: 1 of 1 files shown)


mchurchward:
Please check the main settings page, and  make sure you have the "Allow reports by email" setting checked. This will display the email row on the download.
Screen Shot 2020-07-20 at 2 43 09 PM

adpe:
Hi @mchurchward 

It's not about that. There was a option, so that you can export a column with the `complete` status, and this column was never in the `downloadoptions`. It was created/added when you checked the checkbox.

Here is a screenshot, dated 20180122:
image

HTH,
Adrian

mchurchward:
Huh. Looks like that disappeared with b0b67aeff5c0b5188e089f62cef4f8315081936f.
Pretty sure that was unintentional. Looks like all the code to handle it is still there. I'll get it back.

mchurchward:
Can you rebase on MOODLE_39_STABLE and resubmit this PR?

adpe:
Done @mchurchward 

As I mentioned in the initial PR message, you see where you deleted these lines? I hope this will get back to work soon.

Greets,
Adrian

PR #306, opened 2020-07-15
Save responses when required are missing on page breaks

### LINKED ISSUE #305: Responses are not stored when other required are missing
Hi @mchurchward 

We've encountered following problem. When you are building a questionnaire with page breaks, it will fail to show the `checked` state on the radio-buttons, as these value aren't stored in the `questionnaire_response_bool` or whatever table.

The error message is correct, and tells you, that you must answer the one reuired missing question, but the other correct answered responses aren't anymore checked on this view.

The same behaviour is, on the last page, when you want to submit a questionnaire.

So I think we need to add these missing functions to the `prev`, `next` and `submit` actions on the formdata.

Greets,
Adrian

adpe:
They will saved when the questionnaire has prev, next navigation buttons and when submitting the questionnaire on the last page.

Will resolve #305.

### DIFF
--- a/questionnaire.class.php
+++ b/questionnaire.class.php
@@ -1110,6 +1110,7 @@ class questionnaire {
             if (empty($msg)) {
                 return;
             }
+            $formdata->rid = $this->existing_response_action($formdata, $userid);
         }
 
         if (!empty($formdata->resume) && ($this->resume)) {
@@ -1124,6 +1125,7 @@ class questionnaire {
             $msg = $this->response_check_format($formdata->sec, $formdata);
             if ($msg) {
                 $formdata->next = '';
+                $formdata->rid = $this->existing_response_action($formdata, $userid);
             } else {
                 $nextsec = $this->next_page_action($formdata, $userid);
                 if ($nextsec === false) {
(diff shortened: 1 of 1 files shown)


adpe:
Hi @mchurchward 

Rebase done!

Cheers,
Adrian

adpe:
Hi @mchurchward 

Thanks for your feedback. I'll have again a look. Thanks for the hint - you know your code better than me. ;-)

Greets,
Adrian

mchurchward:
I think the "existing_response_action" would be the function to use. You should also check if this fix is needed for the mobile version.

adpe:
Hi @mchurchward 

This line `response_insert` is already in this function, see https://github.com/PoetOS/moodle-mod_questionnaire/blob/master/questionnaire.class.php#L1041-L1051. The problem is that previous_page_action and next_page_action functions are not called when a requried field is missing, as the function `response_check_format` will return a true as there are missing questions.

Here's a snippet for the previos_page_action, see https://github.com/PoetOS/moodle-mod_questionnaire/blob/master/questionnaire.class.php#L1146-L1158. So actually what I can do is to replace my code with the `existing_response_action` function.

Please have a look at this issue carefully, so that you understand what is wrong.

HTH,
Adrian

mchurchward:
Sorry for the back and forth. Your PR has fallen behind MOODLE_39_STABLE again. Can you rebase with MOODLE_39_STABLE once more? That should make the PR work properly.

mchurchward:
I went ahead and rebased this myself. I'll close it now.

adpe:
Was done in https://github.com/PoetOS/moodle-mod_questionnaire/commit/85b78eed8d9e076b8dabaf34943b5ade5fd15916.

Thank you very much!

PR #323, opened 2020-12-22
PHP 7.4 

ewallah:
In PHP 7.4 there are [Behat exceptions and notices](https://github.com/ewallah/moodle-mod_questionnaire/runs/1593859188?check_suite_focus=true#step:18:125): Trying to access array offset on value of type null in [dirroot]/mod/questionnaire/classes/question/question.php on line 975.  This version of php is stricter than its predecessor.

A good sample is the line: `$respcols->{$colname}->respondentlink[] = $resparr[$a];` Splitting this line in 2 steps assures that the line is not executed from right to left:
```
$respcols->{$colname} = new stdClass();
$respcols->{$colname}->respondentlink[] = $resparr[$a];

```

Another sample is the situation where $form->_customdata is not set.  PHP 7.3 let this pass without warning, PHP 7.4 complains. So you need an additional check to see if $form->_customdata is set:

`$this->form_question_text($mform, isset($form->_customdata) ? $form->_customdata['modcontext'] : '');`

The result of these changes can be seen in [Github actions](https://github.com/ewallah/moodle-mod_questionnaire/runs/1594258869?check_suite_focus=true#step:18:124).  (I also needed to add some changes to a behat script because for some reason we do not arrive at page 1 of the report, but on page 5.)


mchurchward:
Thanks for the contribution, but I'm not seeing those issues in the latest code ([example](https://github.com/PoetOS/moodle-mod_questionnaire/runs/3243226808?check_suite_focus=true#step:17:1)). It may be that it was already fixed.
If you see these again, feel free to open a new issue and submit.

PR #330, opened 2021-01-16
Single choice csv export bug fix

rezeau:
Here's a suggested fix


mchurchward:
Confirming... The bug this is intended to fix is for CSV export only. If a value like:
"2::coffee" is specified for a choice, the CSV export for that choice is expected to be "2". Instead, it is the entire string "2::coffee".

mchurchward:
Currently, the CSV export displays "2 : 2::coffee". The code provided changes that to "2 : 2". That doesn't seem right either?

PR #342, opened 2021-06-18
Stop escaping of text that has been run through the filters as html

andrewhancox:
The mustache variable that I've marked as not needing escaping is a bit of text that gets run through text filters and could end up with html in.

### DIFF
--- a/templates/results_choice.mustache
+++ b/templates/results_choice.mustache
@@ -87,7 +87,7 @@
 {{#responses}}
     {{#response}}
     <tr class="">
-        <td class="cell c0" style="text-align:left;width:50%;">{{response.text}}</td>
+        <td class="cell c0" style="text-align:left;width:50%;">{{{response.text}}}</td>
         <td class="cell c1" style="text-align:left;width:40%;white-space:nowrap;">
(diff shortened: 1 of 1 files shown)


mchurchward:
I don't understand what the problem is that this PR will fix. Can you describe a test case with the problem so that this can be tested for correctness?
Also, can you rebase your branch with the latest MOODLE_310_STABLE code?

andrewhancox:
I've rebased.
The template variable response.text comes from a piece of text that can contain HTML due to filters etc. being applied to it and so needs to be wrapped in {{{ rather than {{ so that the mustache rendering engine knows not to escape the contents.

mchurchward:
So then, this would be for answers where you have provided choice content with HTML in them? Is that right? In which case, can you provide a specific example (unit test, behat test, or steps) to create a question where the results here would be incorrect before the fix?
Thanks.

aruethe2:
I'm encountering this problem with some teachers that are using LaTeX and HTML in their multiple choice answers.  The questionnaire shows up properly for students but the response pages shows escaped HTML instead of LaTeX.    

Example of LaTeX in the questionnaire (works fine):
image

Example of the same questions showing escaped HTML instead of LaTeX in response summary.  
image

I tried adding the extra curly brackets as the the pull requests proposes and it fixed the problem for me.  

The simplest way to verify the issue might be to use an HTML tag in an answer to a multiple choice question.  For example:
H<sub>2</sub>O

The above will render with a subscript "2" in the questionnaire, but will show the html <sub> tags in the response summary.

mchurchward:
I created a Behat test that verifies the problem and the solution, and created a [new branch and PR](https://github.com/PoetOS/moodle-mod_questionnaire/pull/439) from this one.

PR #347, opened 2021-07-07
Questionnaire: improve navigation layout

toanlamt:


toanlamt:
We are from the Open University and would like to do this change: improve navigation layout as:
  - Responsive text input boxes.
  - Next' buttons right, 'Previous' button kept left.
  - Repeat notification/alert at the point of the error, so 'Please answer required question' would appear within the question.

mchurchward:
I rebased and merged this with PR #433.

PR #349, opened 2021-07-09
Support sectiontext question on mobile

andrewhancox:
The section text (label) question is not currently supported on mobile, this PR addresses that.


mchurchward:
Hi. Thank you for this. Is it possible that you could reapply this to the current MOODLE_310_STABLE branch? That branch has the templates for both the latest ionic5 and the previous ionic3.

Ryan-2341:
I can confirm that implementing this solution did work for my team. One (possibly) related issue that has nothing to do with this PR is for a mobile user to access the questionnaire while in Airplane mode (completely offline). Albeit, this PR does allow the user to view labels within the mobile app.

PR #359, opened 2021-09-08
Fixes PHP8 deprecation warnings

mwehr:
Fixes PHP8 deprecation warnings
see https://php.watch/versions/8.0/deprecate-required-param-after-optional


mchurchward:
Thanks for this!

PR #363, opened 2021-10-06
Questionnare: Export support custom fields

toanlamt:
+ Support custom  profile fields in the questionnaire response download summary


emanoylov:
This change has been re-submitted via pull request #536.

Issue #371, opened 2021-11-29
[Moodle 3.9.x] [mod_questionnaire 3.9.1] response "rid" is read before it is ever written to


danowar2k:
It seems that somewhere a rid of a response is read before it is ever written to. Looks like the reading function doesn't know if the response is from the database or not...? (But I don't know what "rid" is exactly, so...)

How to provoke:

* Create a new questionnaire, give it a name, "Save and Display"
* Click on "Add Questions"
* Select "Rate (Scale 1..5)" and "Add select question type"
* Enter a question text and possible answers:

```
angry
calm
hungry
```

* "Save changes"
* Click on "Preview" tab
* Select a value in each line
* Click "Submit Preview"

Output contains:

```
Notice: Undefined property: stdClass::$rid in /var/htdocs/moodle/mod/questionnaire/classes/responsetype/response/response.php on line 116

Notice: Undefined property: stdClass::$rid in /var/htdocs/moodle/mod/questionnaire/classes/responsetype/rank.php on line 69

Notice: Undefined property: stdClass::$rid in /var/htdocs/moodle/mod/questionnaire/classes/responsetype/rank.php on line 69

Notice: Undefined property: stdClass::$rid in /var/htdocs/moodle/mod/questionnaire/classes/responsetype/rank.php on line 69

Notice: Undefined property: stdClass::$rid in /var/htdocs/moodle/mod/questionnaire/classes/responsetype/response/response.php on line 116

Notice: Undefined property: stdClass::$rid in /var/htdocs/moodle/mod/questionnaire/classes/responsetype/rank.php on line 69

Notice: Undefined property: stdClass::$rid in /var/htdocs/moodle/mod/questionnaire/classes/responsetype/rank.php on line 69

Notice: Undefined property: stdClass::$rid in /var/htdocs/moodle/mod/questionnaire/classes/responsetype/rank.php on line 69
```

Questionnaire settings dump:

```
plugin;name;value
"questionnaire";"allowemailreporting";"0"
"questionnaire";"downloadoptions";"response,submitted,institution,department,course,group,id,fullname,username"
"questionnaire";"maxsections";"10"
"questionnaire";"usergraph";"1"
```

danowar2k:
Oh there is a second line in there...I'll do a patch for that one, too.

danowar2k:
I'll look into it...

danowar2k:
[...]

PR #372, opened 2021-11-30
Possible fix for #371

danowar2k:
* Works if you support only PHP 7.0+ (https://www.php.net/manual/de/migration70.new-features.php)
* Works if it doesn't matter if rid is set or not (It probably doesn't matter because otherwise there would be other bugs...)


mchurchward:
This was fixed by other commits in the 3.11 branch.

PR #373, opened 2021-11-30
Patch for response.php, fixes other part of #371

danowar2k:
Just like the first one...


mchurchward:
This was fixed by other commits in 3.11. Closing.

PR #395, opened 2022-04-08
Fix for PHP8; no count() on stdClass

nwp90:
PHP8 does not allow count() on stdClass, so initialise haschildren to null, check for that and use a more complicated test for emptiness on haschildren.

### DIFF
--- a/questions.php
+++ b/questions.php
@@ -69,7 +69,7 @@ if (!$questionnaire->capabilities->editquestions) {
 }
 
 $questionnairehasdependencies = $questionnaire->has_dependencies();
-$haschildren = [];
+$haschildren = null;
 if (!isset($SESSION->questionnaire)) {
     $SESSION->questionnaire = new stdClass();
 }
@@ -187,7 +187,7 @@ if ($action == 'main') {
                 // just remove the dependency and inform the user about it.
                 $haschildren = $questionnaire->get_all_dependants($qid);
             }
-            if (count($haschildren) != 0) {
+            if ($haschildren !== null and (count($haschildren->directs) + count($haschildren->indirects)) > 0) {
                 $action = "confirmdelquestionparent";
             } else {
                 $action = "confirmdelquestion";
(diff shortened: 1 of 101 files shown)


mwehr:
Voting for integration

questions.php
```diff
@@ -187,7 +187,7 @@
                 // just remove the dependency and inform the user about it.
                 $haschildren = $questionnaire->get_all_dependants($qid);
             }
-            if (count($haschildren) != 0) {
+            if ($haschildren !== null and (count($haschildren->directs) + count($haschildren->indirects)) > 0) {
```
mchurchward, line 190: The term "and" should be replaced by "&&".

questions.php
```diff
@@ -69,7 +69,7 @@
 }
 
 $questionnairehasdependencies = $questionnaire->has_dependencies();
-$haschildren = [];
+$haschildren = null;
```
mchurchward, line 72: This line can be deleted completely, in combination with the suggestions below.

questions.php
mchurchward, line 190: The block starting at "if ($questionnairehasdependencies) {" can be simplified down to:

```
            $action = "confirmdelquestion";
            if ($questionnairehasdependencies) {
                // Important: due to possibly multiple parents per question
                // just remove the dependency and inform the user about it.
                $haschildren = $questionnaire->get_all_dependants($qid);
                if (count($haschildren->directs) + count($haschildren->indirects) > 0) {
                    $action = "confirmdelquestionparent";
                }
            }

```
The $haschildren will always be a stdClass with "directs" and "indirects" properties, so we can just check the "questionnairehasdependencies" boolean to determine the rest.

mchurchward (changes requested):
Can you apply the code changes I have highlighted?
I would recommend creating a new branch from MOODLE_311_STABLE and resubmitting a new PR in order to get a clean set of commits.

questions.php
mchurchward: This was already implemented in MOODLE_400_STABLE as part of https://github.com/PoetOS/moodle-mod_questionnaire/pull/442. I will backport it.

PR #396, opened 2022-04-19
Questionnaire: "'Save is actually 'Save and exit', and should be better labelled. Also it should be easier to resume if pressed in error".

lamtranb:


lamtranb:
- Change "Save" button to "Save and exit"
- Add a "Resume" link/button: "Resume Questionnaire".

mchurchward (approved):
This is definitely a good improvement.

PR #401, opened 2022-05-10
Questionnaire: deletion area (Recycle bin) for deleted/orphaned questions

hatuanem199801:
We have to create a new feature the deletion question in mod questionnaire.
Please review this commit and inform for any issues.
Thanks.


mchurchward:
I apologize for the long delay in responding to this. This is an interesting idea. Does this utilize any of the core recycle bin functions?

emhat098:
No, it doesn't. We are using the cron task functions to delete the question at the moment.

tailetan:
This commit is rebased with MOODLE_401_STABLE
cc: @mchurchward

mchurchward:
Hi. This needs another rebasing, and conflict resolution.

tailetan:
I have just rebased with MOODLE_401_STABLE
cc: @mchurchward

mchurchward:
Sorry. The latest branch is the MOODLE_404_STABLE branch. This needs to be rebased against it. I am still trying to understand this. Is the goal to remove questions entirely from the database? In which case, I am not sure that a question that has been flagged as deleted can currently ever be recovered. It may make more sense to simply remove the question at deletion time. Unless there is somewhere that a question can be recovered?

mchurchward:
I still need to understand the workflow of this feature. At present, as far as I know, when a question is deleted, there is no way to actually recover it. It just gets tagged as deleted. Since there is no way to recover it, perhaps it should just be removed from the database at that time. Unless this feature is actually adding a function to recover it?

tailetan:
Hi @mchurchward , I would like to share the ideal workflow for this request as below:
1. When a user creates a question, it appears in the Manage questions section. 
2. If the user deletes the question, a confirmation pop-up will appear stating that the question will be moved to the Question Deletion Area. 
3. If the user clicks "Yes," the question is moved to the Question Deletion Area located below the Manage questions section.
4. Each deleted question will display a message indicating when it will be permanently deleted. 
For example: Each question in the Deletion Area displays a label:
"Time of permanent deletion: night of <deletion time>"
(This time is calculated as: the moment it was moved to the deletion area + the number of days configured in the admin setting. Default: 7 days.)
Eg:
* Moved to deletion area: 30 March 2023
* Admin setting: keep for 7 days
* Time of permanent deletion: night of Wed 06 April 2023
There are three possible outcomes from this point:
1. If the user clicks the X button on the question in the Question Deletion Area, a confirmation pop-up will appear. Clicking "Yes" will permanently delete the question from the system, while clicking "No" will cancel the action.
2. If the user wants to keep the question, they can click the Restore button (arrow icon). The question will then be restored to its original position in the Manage questions section.
3. If the user takes no action, the question will remain in the deletion area until the scheduled task "Empty Questionnaire Recycle Bin" runs. Once the retention period expires, the cron job will permanently delete the question from the database.

Note: The permanent deletion depends on the system’s cron job running as scheduled. Users should pay attention to the "Time of permanent deletion" label if they want to recover a question before it is lost.
 
If you want us to explain more details note with some images/icons please let us know.

mchurchward:
Thanks for your workflow description. Does this change implement that entire workflow? Or is this intended to be one part of it?

tailetan:
@mchurchward yes this change implement that entire workflow.

mchurchward:
Okay, great. I will review this in its entirety.
For now, can you change the version update from 2025060600.00 to 2022121601.02 please. It is easier for me to manage the major version updates at release time.

tailetan:
@mchurchward I have just updated the version. Please review it. Thank you.

db/upgrade.php
```diff
@@ -1016,6 +1016,35 @@ function xmldb_questionnaire_upgrade($oldversion=0) {
...
+            $dbman->change_field_type($table, $field);
+        }
+        unset($field);
+
+        $field = new xmldb_field('deleted', XMLDB_TYPE_INTEGER, '10', XMLDB_UNSIGNED, null, null, null, 'required');
+        if ($dbman->field_exists($table, $field)) {
+            $sql = "UPDATE {questionnaire_question}
+                       SET deleted = ?
+                     WHERE deleted = 'y'";
+            $DB->execute($sql, [time()]);
```
mchurchward: This should not need a $DB->execute. A $DB->set_field() should be used.

db/upgrade.php
```diff
@@ -1016,6 +1016,35 @@ function xmldb_questionnaire_upgrade($oldversion=0) {
...
+            $sql = "UPDATE {questionnaire_question}
+                       SET deleted = null
+                     WHERE deleted = 'n'";
+            $DB->execute($sql);
```
mchurchward: This should not need a $DB->execute. A $DB->set_field() should be used.

db/upgrade.php
```diff
@@ -1016,6 +1016,35 @@ function xmldb_questionnaire_upgrade($oldversion=0) {
...
+        }
+        $field = new xmldb_field('deleted', XMLDB_TYPE_CHAR, '10', XMLDB_UNSIGNED, null, null, null, 'required');
+        if ($dbman->field_exists($table, $field)) {
+            $dbman->change_field_type($table, $field);
+        }
+        unset($field);
+
+        $field = new xmldb_field('deleted', XMLDB_TYPE_INTEGER, '10', XMLDB_UNSIGNED, null, null, null, 'required');
+        if ($dbman->field_exists($table, $field)) {
+            $sql = "UPDATE {questionnaire_question}
```
mchurchward: Have these update steps been tested on large data tables? Many sites have an extremely large number of questions.
Since existing deleted questions have not been restorable prior to this change, I think it makes more sense to simply delete the records during the upgrade step. If it is set to a time like this step does now, I think the first run of the cron task could easily hang and easily run out of memory.

locallib.php
```diff
@@ -300,6 +303,85 @@ function questionnaire_delete_survey($sid, $questionnaireid) {
...
+}
+
+/**
+ * Restore deleted questions.
+ *
+ * @param int $qid question id.
+ * @param int $sid survey id.
+ * @return void
+ */
+function questionnaire_restore_deleted_question($qid, $sid) {
```
mchurchward, line 348: Is this function tested well? Especially with multiple positions?

locallib.php
```diff
@@ -300,6 +303,85 @@ function questionnaire_delete_survey($sid, $questionnaireid) {
...
+        $updatesql = "UPDATE {questionnaire_question}
+                         SET deleted = null, position = :position
+                       WHERE id = :qid
+                         AND surveyid = :sid";
+        $params = [
+                'position' => $question->lastposition ?? 1,
+                'qid' => $qid,
+                'sid' => $sid
+        ];
+        $DB->execute($updatesql, $params);
```
mchurchward: This should use $DB->update_record(), not $DB->execute(). Is there ever a condition where 'sid' is needed?

locallib.php
```diff
@@ -951,6 +1055,27 @@ function questionnaire_get_standard_page_items($id = null, $a = null) {
...
+ * @param int $qid question id.
+ * @param int $qtype question type.
+ * @return int number or 0 if responses were not found.
+ */
+function count_reponses_question($qid, $qtype) {
+    global $DB;
+
+    $countresps = 0;
+    if ($qtype != QUESSECTIONTEXT) {
+        $responsetable = $DB->get_field('questionnaire_question_type', 'response_table', array('typeid' => $qtype));
```
mchurchward: array() should use the [] construct. Same with below.

questions.php
```diff
@@ -113,28 +127,49 @@
...
     }
     $reload = true;
 }
 
+// Delete question permanently.
+if ($delpermanentlyq) {
+    $qid = $delpermanentlyq;
+    $sid = $questionnaire->survey->id;
+    questionnaire_delete_permanently_questions($qid, $sid);
+    $deletedQuestion = $questionnaire->deletequestions[$qid] ?? null;
```
mchurchward: Should not be camelCase

mchurchward (changes requested):
I have left some inline comments that need to get dealt with. I will continue the functional review as well.

When I upgrade from the previous version of questionnaire, and go to the "Question deletion area" of the edit questions function, the new icons are huge and not usable.

There is a [failing Behat test](https://github.com/PoetOS/moodle-mod_questionnaire/actions/runs/18981317508/job/54214329588#step:16:215).

mchurchward:
I have rebased it and merged it in with [pull request 676](https://github.com/PoetOS/moodle-mod_questionnaire/pull/676).

PR #403, opened 2022-05-12
Questionnaire: new 'Slider' type question

toanlamt:


toanlamt:
+ Create a new question type Slider
Please review this commit and inform for any issue
Thanks.

mchurchward:
Thanks... Some comments/questions:

1. I'm not sure this needs a new response table. This looks like it was one text based response for every question, so couldn't this just use the existing "questionnaire_response_text" table? This is what the numerical question does.
2. Can you provide the mobile templates and any code as well?

toanlamt:
Hi @mchurchward 
Sorry it took so long to respond. I have solved the issues you raised in the previous comment. Could you please help me code review again?
Many Thanks

db/mobile.php
```diff
@@ -35,9 +35,9 @@
                 'delegate' => 'CoreCourseModuleDelegate',
                 'method' => 'mobile_view_activity',
                 'styles' => [
-                    'url' => $CFG->wwwroot . '/mod/questionnaire/styles_app.css',
-                    'version' => '1.4'
-                ]
+                        'url' => $CFG->wwwroot . '/mod/questionnaire/mobile_css.php?v=2022092200',
```
mchurchward, line 38: What is this change for? I'm not familiar with this.

classes/questions_form.php
```diff
@@ -256,7 +256,7 @@ public function definition() {
                 $manageqgroup[] =& $mform->createElement('image', 'editbutton['.$question->id.']', $esrc, $eextra);
                 $manageqgroup[] =& $mform->createElement('image', 'removebutton['.$question->id.']', $rsrc, $rextra);
 
-                if ($tid != QUESPAGEBREAK && $tid != QUESSECTIONTEXT) {
+                if ($tid != QUESPAGEBREAK && $tid != QUESSECTIONTEXT  && $tid != QUESSLIDER) {
```
mchurchward, line 259: Why can't a slider question be required?

classes/responsetype/slider.php
```diff
@@ -0,0 +1,54 @@
...
+class slider extends numericaltext {
+    /**
+     * Return an array of answer objects by question for the given response id.
+     * THIS SHOULD REPLACE response_select.
+     *
+     * @param int $rid The response id.
+     * @return array array answer
+     * @throws \dml_exception
+     */
+    public static function response_answers_by_question($rid) {
```
mchurchward, line 36: Why would the standard text parent function not work here? What is special about the responses for slider?

mod/questionnaire/mobile_css.php
```diff
@@ -0,0 +1,31 @@
+<?php
```
mchurchward, line 1: Not sure what this file is. Can you explain?

mchurchward (commented):
I have left several questions. Please respond at you earliest convenience.

db/mobile.php
toanlamt: We will revert this change and just adjust the version. Please see the last comment for more detail.

classes/questions_form.php
toanlamt: Normally the slider question will have a default value, the user can answer or keep it, so the required condition here is not necessary.

classes/responsetype/slider.php
toanlamt: We need to get an additional "extradata" field from 'questionnare_question' table for response_slider.mustache template. 
So the response will show a slider with the value selected by the user not just plain text like question_text

mod/questionnaire/mobile_css.php
toanlamt: Sorry for confusing you. In our project, we want to allow CSS files can be accessed from external requests, so we add some headers.
This is specific to our project so we will revert this change.

toanlamt:
Hi @mchurchward
Thank you for your feedback. I added some comments, could you please help me to take a look?
If you have any questions, please let me know.

classes/questions_form.php
mchurchward: Okay. I will test this that way.

version.php
```diff
@@ -25,7 +25,7 @@
 
 defined('MOODLE_INTERNAL') || die();
 
-$plugin->version  = 2022030301;  // The current module version (Date: YYYYMMDDXX).
+$plugin->version  = 2022092200;  // The current module version (Date: YYYYMMDDXX).
```
mchurchward, line 28: I apologize that I didn't notice this before, but the version bump needs to retain the '20220303' portion. This date portion represents the MOODLE_400_STABLE branch.
The MOODLE_401_STABLE branch, if it becomes needed, will have a greater than '20220303' portion.
I will change this. I hope it doesn't negatively impact your installs. If it does, let me know, and I'll release a 4.01 branch with a higher version.

mchurchward (approved):
This is a great contribution and addition. Thanks!

version.php
mchurchward: I take it back. Since I haven't released MOODLE_400_STABLE yet, I'll just adopt the '20220922' as the base version number for this branch.

Issue #415, opened 2022-08-09
Conditional Branches, Rating scales and Required questions


rezeau:
In a 'branching questionnaire', rating scale questions set as 'required', can be be skipped! This is a bug.
See https://moodle.org/mod/forum/discuss.php?d=437073#p1759040

danowar2k:
@mchurchward Could you please verify if this fix is correct without side effects? We'd then incorporate it into our version until we switch to a fixed questionnaire version...

mchurchward:
This has been fixed as above.

PR #416, opened 2022-08-09
fix issue #415

rezeau:
Signed-off-by: rezeau <joseph@rezeau.org>


mchurchward:
Replacing with https://github.com/PoetOS/moodle-mod_questionnaire/pull/446

PR #422, opened 2022-09-16
Error when date does not come as string in PSQL

lucaboesch:
When a field is a text field and another field is a date field, when downloading CSV (or having mail activated and submit), on PosgreSQL, there is a ERROR: UNION type text and date cannot be matched.

### DIFF
--- a/classes/responsetype/date.php
+++ b/classes/responsetype/date.php
@@ -283,4 +283,28 @@ class date extends responsetype {
     protected function bulk_sql_config() {
         return new bulk_sql_config(static::response_table(), 'qrd', false, true, false);
     }
+
+    /**
+     * Return sql for getting responses in bulk.
+     * @return string
+     */
+    protected function bulk_sql() {
+        global $DB;
+
+        $userfields = $this->user_fields_sql();
+        $alias = 'qrd';
+        if ($DB->get_dbfamily() === 'postgres') {
+            $extraselect = "0 AS choice_id, TO_CHAR(qrd.response::date, 'mm-dd-yyyy') AS response, 0 AS rankvalue";
+        } else {
+            $extraselect = "0 AS choice_id, qrd.response AS response, 0 AS rankvalue";
+        }
+
+        return "
+            SELECT " . $DB->sql_concat_join("'_'", ['qr.id', "'".$this->question->helpname()."'", $alias.'.id']) . " AS id,
+                   qr.submitted, qr.complete, qr.grade, qr.userid, $userfields, qr.id AS rid, $alias.question_id,
+                   $extraselect
+              FROM {questionnaire_response} qr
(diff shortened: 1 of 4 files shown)


mchurchward (commented):
I'm not sure why this would be needed. The "response" field in the "questionnaire_response_date" table is a text field as are the other 'response' fields in the other response tables. Where is the "date" field that needs converting?

lucaboesch:
image004
Dear @mchurchward 
This is the error triggered otherwise.
If you wish I can write a Behat test.
Best,
Luca

lucaboesch:
Reproduction steps:
- on a PostgreSQL Moodle, as a teacher, insert a questionnaire and add a text question and a date question.
- as student, fill in an answer.
- as teacher, try to download the results.

mchurchward:
Please review the expected data type for the "[questionnaire_response_date:response](https://github.com/PoetOS/moodle-mod_questionnaire/blob/MOODLE_400_STABLE/db/install.xml#L147)" field on the site you are having the issue with. This should be a text field as are all other "response" fields in that SQL statement. The error you are saying is happening would be caused by the "questionnaire_response_date:response" field being set to a "date" type.
Prior to version 2016020204 (Moodle 3.1), this "may" have been a "date" type, but has been a text field ever since then.

classes/responsetype/date.php
```diff
@@ -283,4 +283,28 @@ public static function response_answers_by_question($rid) {
     protected function bulk_sql_config() {
         return new bulk_sql_config(static::response_table(), 'qrd', false, true, false);
     }
+
```
mchurchward, line 286: This change is not needed. The date value is already a text field.

mchurchward (commented):
Hi Luca. I [made a copy of your branch, removed the changes to the date class](https://github.com/PoetOS/moodle-mod_questionnaire/commit/91e0a9b438806c20eff063be29dc3257f53a018f), and ran the [CI with your Behat test](https://github.com/PoetOS/moodle-mod_questionnaire/actions/runs/3759537293/jobs/6389219164). The test passed using Postgres and MySQL. There really should be no reason for this change.
That said, thanks for pointing out the "label" issues with the mustache and rendering code. This is an accessibility issue that I will deal with separately.

lucaboesch:
Dear @mchurchward 

You are absolutely right. I rest my case, the reason why this was a thing is because the field still was of type date.

```

   Column    |  Type   | Collation | Nullable |                           Default                           
-------------+---------+-----------+----------+-------------------------------------------------------------
 id          | integer |           | not null | nextval('mdl_questionnaire_response_date_id_seq'::regclass)
 response_id | bigint  |           | not null | 0
 question_id | bigint  |           | not null | 0
 response    | date    |           |          | 

```

I close this pull request.

Best,
Luca

PR #429, opened 2022-10-14
Fix feedback section errors when scorecalculation is empty

jason-platts:
Fixes array_key_exists errors in label text when the scorecalculation field is empty string and not array.

This happens when:

1. You select 'feedback sections' option in the 'feedback' tab
2. You select 'Save settings and edit Feedback Sections'
3. You don't then save on the resulting feedback sections screen

Also sets the initial feedback section DB record to have empty serialised array instead of empty string so data matches completed records

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,26 +8,16 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - php: '7.4'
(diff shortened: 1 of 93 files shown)


mchurchward (approved):
Wondering if section::decode_scorecalculation should be refactored and used in sectiontext.php?
Approving without this.

PR #430, opened 2022-10-18
M4: Warning displayed when editing the completion tracking option

tailetan:
Fix warning message displayed when editing the completion tracking option

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -240,6 +240,36 @@ function questionnaire_delete_instance($id) {
 }
 
 /**
+ * Add a get_coursemodule_info function in case any questionnaire type wants to add 'extra' information
+ * for the course (see resource).
+ *
+ * Given a course_module object, this function returns any "extra" information that may be needed
+ * when printing this activity in a course listing.  See get_array_of_activities() in course/lib.php.
+ *
+ * @param stdClass $coursemodule The coursemodule object (record).
+ * @return cached_cm_info An object on information that the courses
+ *                        will know about (most noticeably, an icon).
+ */
+function questionnaire_get_coursemodule_info($coursemodule) {
+    global $DB;
+
+    $questionnaire = $DB->get_record('questionnaire',
+        array('id' => $coursemodule->instance), 'id, name, intro, introformat,
+             completionsubmit');
+    if (!$questionnaire) {
+        return null;
+    }
+
+    $info = new cached_cm_info();
+    $info->customdata = (object)[];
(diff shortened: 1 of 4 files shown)


mchurchward (commented):
Can I get a better description of what this does? Is this standard Moodle or custom work from the OU?

tailetan:
> Can I get a better description of what this does? Is this standard Moodle or custom work from the OU?

Hi Mike, 
Sorry for the unclear description. This is Moodle standard of custom completion. It requires a custom completion class

For more detail, please check this documentation https://docs.moodle.org/dev/Student_activity_completion#Custom_activity_completion

Thank you.

lib.php
```diff
@@ -1185,15 +1215,13 @@ function questionnaire_reset_userdata($data) {
...
- * @param stdClass $cm Course-module
+ * @param object $cm Course-module
  * @param int $userid User ID
  * @param bool $type Type of comparison (or/and; can be used as return value if no conditions)
  * @return bool True if completed, false if not, $type if conditions not set.
  *
- * $course is unused, but API requires it. Suppress PHPMD warning.
  */
-function questionnaire_get_completion_state($course, $cm, $userid, $type) {
+function questionnaire_get_completion_state_lib($cm, $userid, $type) {
```
mchurchward: Why has this been renamed? Since the callback function is now replaced with the "completion\custom_completion" class, shouldn't this be removed? Does renaming it allow it to be used somewhere else?

mchurchward (commented):
Thanks for this. If you can just clarify the one change in lib.php, we're good to go. Thanks fr the unit and Behat tests as well.

tailetan:
Hi @mchurchward ,
Could you please review my newest update please?
Thank you

mchurchward:
From what I can tell, the only difference in your latest commit is changing
`function questionnaire_get_completion_state_lib($cm, $userid, $type) {`
to
`function questionnaire_get_completion_state($cm, $userid, $type) {`

But, is this function even needed any more?

tailetan:
> From what I can tell, the only difference in your latest commit is changing `function questionnaire_get_completion_state_lib($cm, $userid, $type) {` to `function questionnaire_get_completion_state($cm, $userid, $type) {`
> 
> But, is this function even needed any more?

Hi,
This function is used in classes/completion/custom_completion.php - line 49.
Thank you.

mchurchward (approved):

Issue #381, opened 2022-01-13
'Non-respondents' tab is broken when nobody responds question.


nhatbao47:
When editor edits a course and add questionnaire activity. They navigate to questionnaire and add new question. After that, they switch to the 'Non-respondents' tab to view incomplete users. The out of memory error happens. Because it returns all users without pagination. The page renders a lot data so it causes out of memory.

mchurchward:
This has now been fixed in MOODLE_311_STABLE and MOODLE_400_STABLE.

PR #437, opened 2022-11-24
#1 - Fixed non respondents tab getting to many users

### LINKED ISSUE #216: Non-respondents list showing every user on system level.
Non-respondents list showing every user on system level who has mod/questionnaire:submit Allowed, even though the users not enrolled on the course. Better would be to show students only who are enrolled on the course.

frederikmillingpytlick:
See https://moodle.org/mod/forum/discuss.php?d=275297

Would probably fix #216 and #381


mchurchward:
Going to close this one. #436 supports back to Moodle 3.9.

PR #441, opened 2022-12-12
Open and close date in course overview

aczerzuh:
Hi,
we've added the feature to display the open and close date for the activity in the course overview. If you need any changes to the code please let me know.
best wishes,
adrian

### DIFF
--- a/classes/dates.php
+++ b/classes/dates.php
@@ -1,0 +1,71 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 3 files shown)


mchurchward (commented):
Please fix the [CI issues](https://github.com/PoetOS/moodle-mod_questionnaire/actions/runs/3675371623/jobs/6357651489#step:10:15).

Also, I'm unclear on what the new "dates" class does. You have added it, but I do not see it being used anywhere?

aczerzuh:
fixed the errors.

The dates class is used by the moodle core. It preps the dates to be displayed in the course view. In this case it is used to determine the label for the dates. 

You could use it to display all kind of dates. A very similar usage is in the quiz module and a different implementation is in the forum module. Although the difference isn't to big.

mchurchward:
Thanks for your responses...
Are you able to add a Behat test for this? It would be good to have that to avoid future regressions.
Also, is this something that would apply to 3.11 as well?

aczerzuh:
Sorry for the delay and the mess with the worflow runs. I rebased the code, changed some lines to work with the new base and added the behat test.
The activity dates can also be displayed in Moodle 3.11, but the support for this version ends pretty soon.
best wishes

rezeau (approved):
works as expected

PR #442, opened 2022-12-12
Questionnaire: Error when trying to delete a question

emanoylov:
Fixes for the following errors when deleting questions or page breaks:

Error 1:

Exception - count(): Argument #1 ($value) must be of type Countable|array, stdClass given

Error code: generalexceptionmessage

line 190 of /mod/questionnaire/questions.php: TypeError thrown
line 190 of /mod/questionnaire/questions.php: call to count()

Error 2:

Exception - count(): Argument #1 ($value) must be of type Countable|array, null given

Error code: generalexceptionmessage

line 799 of /mod/questionnaire/locallib.php: TypeError thrown
line 799 of /mod/questionnaire/locallib.php: call to count()
line 133 of /mod/questionnaire/questions.php: call to questionnaire_check_page_breaks()


jason-platts (approved):
Thanks, it looks like these changes will fix the errors

PR #446, opened 2022-12-20
Rezeau fix issue 415

mchurchward:


mchurchward:
Reviewed the original PR, but needed rebasing, so the original commit was cherry-picked here.

PR #455, opened 2023-01-08
Questionnaire: Error adding/deleting question with branch/no branch

emanoylov:
Fixes for various errors and warnings:

PHP Warning:  count(): Parameter must be an array or an object that implements Countable in /www/html/mod/questionnaire/locallib.php on line 765

PHP Warning: Invalid argument supplied for foreach() in /var/www/html/moodle/mod/questionnaire/locallib.php on line 754

PHP Notice:  Trying to access array offset on value of type null in /var/www/html/moodle/mod/questionnaire/locallib.php on line 751

PHP Warning:  count(): Parameter must be an array or an object that implements Countable in /www/html/mod/questionnaire/locallib.php on line 746

PHP Warning:  Invalid argument supplied for foreach() in /www/html/mod/questionnaire/locallib.php on line 748

PHP Notice:  Undefined offset: 11 in /www/html/mod/questionnaire/locallib.php on line 710

### DIFF
--- a/classes/feedback/section.php
+++ b/classes/feedback/section.php
@@ -219,13 +219,14 @@ class section {
         $DB->delete_records(self::TABLE, ['id' => $this->id]);
 
         // Resequence the section numbers as necessary.
-        $allsections = $DB->get_records(self::TABLE, ['surveyid' => $this->surveyid], 'section ASC');
-        $count = 1;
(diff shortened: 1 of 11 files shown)


jason-platts (approved):
Thanks.

We've also tested manually and checked automated tests here so I think this change is safe to merge.

PR #462, opened 2023-01-31
CONTRIB-9171 Safely decode score calculation field

mchurchward:


mchurchward:
Reviewed on [Moodle tracker](https://tracker.moodle.org/browse/CONTRIB-9171).

PR #464, opened 2023-02-01
Ghi456 311

mchurchward:


basilevm:
Tested updated code and no longer receiving error: Exception - Attempt to modify property "data" on null

PR #467, opened 2023-02-16
Fix radio group percentage

zefoudendorp:
Fix the total number of respondents for radiobutton questions in groups.


mchurchward:
Hi. Can you describe the problem and the solution please? If there is a reference issue, then just link it. Thanks.

PR #482, opened 2023-05-16
Update sectiontext questions to display on mobile

sayoder-aprende:
Also refactored the code a bit for clarity: The question class now has an is_numbered() method (default true) that can be overridden for question types that should not have numbers. Replaced some checks for specific question types with checks for is_numbered().


mchurchward:
This has broken a number of Behat tests. Can you review and correct as needed?

sayoder-aprende:
Sorry about that, I got caught up in something else and neglected to come back and check whether the test runners succeeded. I will try to get a build out that passes CI this week!

sayoder-aprende:
Hi again,

The test runner logs have been deleted since I took so long to address this (and my local PHPUnit shows no issues), so I'm going to close this PR and create a new one so I can see what went wrong.

PR #488, opened 2023-06-15
Questionnaire: More meaningful error than nopermissions

emanoylov:
Defines the nopermissions string in the questionnaire plugin.


mchurchward (approved):
Thanks. Looks good.

PR #495, opened 2023-07-10
Questionnaire\Accessibility\Slider: The slider values should be associated with the labels and the slider should be programmatically associated with the question

tailetan:
Hi @mchurchward ,

We are from the Open University and would like to propose the following changes:

1. Fix minor errors in template files to ensure that the screen reader can read the entire question.
2. Implement a new function to support accessibility screen readers. Enable the screen reader to read out the value in context when the user changes the slider value and there are corresponding labels.

Here are some examples of the desired output:

If the slider value is [Slider value], the labels would be [left label], [centre value], and [right label].
For example, if the labels are "[1|2|3|4|5|6|7|8|9]", where 1 represents 'strongly disagree' and 9 represents 'strongly agree'.
In another case where all points have labels, the format would be "[1|2|3|4|5|6|7|8|9]". For instance, 1 stands for 'strongly disagree', 5 stands for 'neither agree nor disagree', and 9 stands for 'strongly agree'.
If there is a scenario where the centre label is shared by two values, such as a scale from 1 to 10 where 5 and 6 represent the centre values, the screen reader should mention that specifically as follows: "5 and 6 represent 'neither agree nor disagree'".

Could you please review this commit and let us know if you have any concerns?

Thank you.


lang/en/questionnaire.php
rezeau, line 1: Tested this PR with the NVDA screen reader. Almost works as expected, except that the "where" string is read letter by letter instead of as a full word. For example NVDA says: 
"5 w h e r e 1 is strongly disagree, 5 and 6 are neither agree nor disagree and 10 is strongly agree"
instead of  "5 where 1 is strongly disagree, 5 and 6 are neither agree nor disagree and 10 is strongly agree"
Easy fix suggested: change $string['where'] = 'where '; to $string['where'] = '- where ';
i.e. add a dash and space.
Edit.- More details. Tested under Windows with NVDA. Browsers: Firefox and MS Edge, the problem (and the suggested solution) are identical in both browsers.
Neither NDVA nor ChromeVox work correctly in the Chrome browser.

tailetan: hi @rezeau,
I have tested this with the NVDA screen reader (Firefox and MS Edge), and it works as expected. Please refer to the evidence in the drive link below. Thank you.
https://drive.google.com/drive/folders/1I_9GVUTux5Hxy-PVTcg6H4i0G0pO75p5?usp=sharing

rezeau: @tailetan Thanks for responding! Whether it works seems to depend upon the "voice" you select in NVDA speech settings! 
Problem shown in linked video happens with those settings:
Synthesizer Microsoft Speech API version 5
Voice: Microsoft Zira Desktop - English (United States)
https://drive.google.com/file/d/10doBFv73nEIqvGYMrNe2pni_hUo6JCFs/view
You can hear "3 w h e r e 1 is never" letter by letter !
No problems with other selections. Which NVDA settings did you use in your drive link test?

tailetan: @rezeau I've noticed the issue. I'll work on resolving it and provide you with an update as soon as possible. Thank you.

tailetan:
hi @rezeau ,
I have updated  the code to fix the NVDA screen reader issue. Could you kindly have a look on this again? Thank you

rezeau:
@tailetan Just tested under windows with NVDA speech settings:
Synthesizer Microsoft Speech API version 5
Voice: Microsoft Zira Desktop - English (United States)
Works as expected!

version.php
mchurchward, line 1: I have changed the base branch for this to MOODLE_401_STABLE. This has likely created a conflict in the version file. I realize you want this bumped due to the js and lang changes. Can you use a decimal bump instead? So from 2022121600 to 2022121600.01?

mchurchward (changes requested):
Can you update the slide behat tests to refect your changes?

version.php
mchurchward: Sorry. One more. Please don't change the $plugin->requires or the $plugin->release lines. The 'requires' may be changed if we think this release will work with 4.0, but that would require another issue. The 'release' should remain the same until we release a new version.

tailetan:
I have rebased and updated some Behat tests. Could you please take a look at this, @mchurchward ?

mchurchward (approved):

PR #496, opened 2023-07-14
Questionnaire\Accessibility\Numeric: Instructions not programmatically associated with field

tailetan:
Hi @mchurchward,

We are from the Open University and would like to propose the change that support the accessibility screen reader to read out the instruction of the Numeric questions.
Could you please review this commit and let us know if you have any concerns?

Thank you indeed.

### DIFF
--- a/templates/question_numeric.mustache
+++ b/templates/question_numeric.mustache
@@ -29,23 +29,37 @@
 
     Example context (json):
     {
-        "qelements": {
-            "choice": {
-                "onkeypress": "dosomething()",
-                "size": 30,
-                "name": "choiceid1",
-                "maxlength": 50,
-                "value": "473",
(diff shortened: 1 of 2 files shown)


rezeau (approved):
Tested with Windows NVDA on Firefox and MS Bing browsers.
Works as expected.

mchurchward (changes requested):
Can you rebase this with MOODLE_401_STABLE?
I don't think the ".idea" directory should be part of this.
Is there a way to update the Behat test to verify this?

tailetan:
I have updated behat test and rebased this with MOODLE_401_STABLE. Could you merge this commit, @mchurchward ? Thanks

mchurchward (approved):

PR #497, opened 2023-07-14
Questionnaire\Accessibility\Rate: table does not have a programmatically associated caption

lamtranb:
Hi @mchurchward 

We are from the Open University and would like to propose the change that support the accessibility screen reader for Rate question

Data tables SHOULD have a programmatically associated caption or name. Screen readers read the caption or name of the table when users navigate to the table, giving users a sense of what the table is about. The caption element is the most straightforward way to give a name to a table. The recommendation is to add an appropriate caption to the table using the <caption> tag with question text.

Could you please review this commit ?

Many thanks.

### DIFF
--- a/classes/question/rate.php
+++ b/classes/question/rate.php
@@ -198,6 +198,7 @@ class rate extends question {
     protected function question_survey_display($response, $descendantsdata, $blankquestionnaire=false) {
         $choicetags = new \stdClass();
         $choicetags->qelements = [];
+        $choicetags->qelements['caption'] = strip_tags($this->content);
 
         $disabled = '';
         if ($blankquestionnaire) {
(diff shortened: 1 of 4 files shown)


rezeau (approved):
Tested with Windows NVDA on Firefox and MS Bing browsers.
Works as expected.

mchurchward (changes requested):
Code changes look good.
Can you rebase and resubmit with MOODLE_401_STABLE?
Is there a way to update the Behat test to verify this change?

lamtranb:
Hi @mchurchward,

I will rebase and resubmit with MOODLE_401_STABLE.
For behat test for accessibility, I think we can use xpath to check if elements exist or elements have correct attributes. I will add behat tests as your suggestion. If you have any better idea, please let me know.

Thanks.

mchurchward (approved):

PR #498, opened 2023-07-28
Fix PHP error message  Warning: Undefined variable $sectionlabel in C…

rezeau:
…:\wamp64\www\htdocs\moodle\mod\questionnaire\questionnaire.class.php on line 4018

### DIFF
--- a/questionnaire.class.php
+++ b/questionnaire.class.php
@@ -4021,8 +4021,7 @@ class questionnaire {
                     $allresponses,
                     $this->survey->chart_type,
                     array_values($scorepercent),
-                    array_values($allscorepercent),
-                    $sectionlabel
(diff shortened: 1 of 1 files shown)


rezeau:
This will fix issue #500

mchurchward (changes requested):
Hi. Sorry for the delay. I am concerned about removing the $sectionlabel, as it is defined in that function, and so may be valid in some cases. Perhaps it just needs to be initialized to tne empty string  to solve this?
Also, can you rebase/reapply against MOODLE_401_STABLE now?

lucaboesch:
#542 will also fix this.

mchurchward:
Since #542  fixed this, I will close this PR.

PR #501, opened 2023-08-01
Questionnaire/Accessibility: Check boxes missing group label

emanoylov:
Changes related to accessibility so that the checkbox answers are better-associated with the question text asked.

E.g. if the question is "Select favourite colours" and the options are Blue/Red/Green the possible responses are read out by the screen reader as:

"Select favourite colours Blue"
"Select favourite colours Red"
"Select favourite colours Green"

Additional related changes:
The question container is modified to add a new qlegend field to the legend and to make the legends consistent as some were missing the qnum field.
The response templates are also modified for consistency.

### DIFF
--- a/classes/question/check.php
+++ b/classes/question/check.php
@@ -81,7 +81,7 @@ class check extends question {
      * @return \stdClass The check question context tags.
      *
      */
-    protected function question_survey_display($response, $dependants, $blankquestionnaire=false) {
+    protected function question_survey_display($response, $dependants, $blankquestionnaire = false) {
         // Check boxes.
         $otherempty = false;
         if (!empty($response)) {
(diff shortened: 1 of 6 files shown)


rezeau:
Tested under Windows with NVDA screenreader and Firefox and MS Edge browsers.
This works as expected. However it introduces changes in a large number of files, so not sure it will be approved.

**Note**. The example provided does not make much sense with check boxes. 
A better use case example would be something like:
What are you favorite drinks?
Coffee
Tea
Wine

mchurchward (changes requested):
Per Joseph's comment, I think this could be done in a much simpler way, that should primarily impact just the question types in question, without the need for passing the $pagetags around all of the functions. Definitely needed though.

emanoylov:
I've minimised the changes to the check box question type as per the comments. Please review again.

mchurchward (commented):
I'm not clear on why the main question file essay question section, and the question and response container templates were modified. Can you explain?

emanoylov:
@mchurchward The changes are after the QUESESSAY if statement; it just splits the formatting of $content to avoid unnecessary formatting for qlegend.
The question container is modified to add qlegend to the legend and to make the legends consistent as some were missing qnum.
The response templates are modified for consistency.

mchurchward:
> The changes are after the QUESESSAY if statement

Ah yes. Sorry, I read that wrong.

> The question container is modified to add qlegend to the legend and to make the legends consistent as some were missing qnum.

Okay. So then this fix is for more that checkbox questions? Is that correct? If so, can you update the description to reflect this?

emanoylov:
@mchurchward  I've updated the description. Please progress the pull request. Thanks.

mchurchward (approved):

PR #505, opened 2023-08-15
Questionnaire\Accessibility\Rate: The form controls within the table are not accessible

lamtranb:
Hi,
We are from The Open University,

As a user, when I use the screen reader tool and access the Rate question on the questionnaire activity
I would probably expect the screenreader to read something like:

Row {number of row}, {name of answer title}: Column {number of column}, {name of answer choice}.

And especially for the first column: {name of answer choice} = Unanswered

The number of column and row should start at 2.

image

Could you please review this change ?

Thanks


mchurchward (changes requested):
Things look good. Can you rebase against MOODLE_401_STABLE and resubmit?

tailetan:
I have rebased this with MOODLE_401_STABLE.
FYI: @mchurchward

mchurchward (approved):

PR #506, opened 2023-08-16
Questionnaire: improved accessibility for essay box type

toanlamt:
Hi,
We are from The Open University,

As a user, when using the screen reader tool to access the Essay box question within the questionnaire activity, we would likely expect the screen reader to read the question.

Could you please review this change?

Many Thanks


mchurchward (changes requested):
Thanks for this. All looks good.
Can you rebase / reapply against the MOODLE_401_STABLE and resubmit?

mchurchward (approved):

Issue #507, opened 2023-09-12
Invalid $required parameter value: '1' . It must be either VALUE_DEFAULT, VALUE_REQUIRED, or VALUE_OPTIONAL


opitz:
PHP 8.0 throws a warning as a parameter does not follow coding standards

To fix this line 133 of mod/questionnaire/externallib.php needs to change from

**'submitted' => new \external_value(PARAM_BOOL, 'submitted', true, false, false),**

to

**'submitted' => new \external_value(PARAM_BOOL, 'submitted', VALUE_REQUIRED, false, false),**

Issue #502, opened 2023-08-01
External API error


aspark21:
Moodle: MOODLE_402_STABLE
Questionnaire: MOODLE_400_STABLE

Visit your.moodle.site/admin/webservice/documentation.php

<img width="1367" alt="Screenshot 2023-08-01 at 16 40 25" src="https://github.com/PoetOS/moodle-mod_questionnaire/assets/4015496/92ac380e-fdf9-4225-9d88-479272f60915">

mchurchward:
I can't replicate this issue. Is there a specific version of PHP required as well?

PR #508, opened 2023-09-12
Fixing issues #502 and #507

opitz:
Using core_external/external_api
Using valid VALUE_REQUIRED parameter


mchurchward (approved):
Not sure the namespace changes were absolutely necessary, but I'll accept them.

PR #490, opened 2023-06-23
New attempt at restoring the radar chart_type to questionnaire feedback

rezeau:
This is my 3rd attempt at restoring this really needed chart type. Don't know why I closed my similar PR. I do hope this new PR can be merged any time soon. It's only a few lines changed in the code; it used to work when I first created it years ago and it disappeared from the code some time ago, no idea why.
Thanks in advance, 
Joseph

PR #510, opened 2023-09-26
restore_radar_charttype rebased to MOODLE_401_STABLE

rezeau:
This PR replaces obsolete PR #490

### DIFF
--- a/classes/feedback_form.php
+++ b/classes/feedback_form.php
@@ -39,10 +39,7 @@ class feedback_form extends \moodleform {
         // We need to get the number of current feedbacksections to allow the rose and radar chart types or not.
         global $DB;
         $sid = $questionnaire->survey->id;
-        $sql = 'SELECT COUNT(*) ' .
-           'FROM {questionnaire_fb_sections} ' .
-           'WHERE surveyid = '.$sid;
-        $numfeedbacksections = $DB->count_records_sql($sql, [$sid]);
+        $numfeedbacksections = $DB->count_records('questionnaire_fb_sections', ['surveyid' => $sid]);
 
         // Get all questions that are valid feedback questions.
         $nbvalidquestions = 0;
--- a/classes/feedback_form.php
+++ b/classes/feedback_form.php
@@ -36,6 +36,21 @@ class feedback_form extends \moodleform {
      */
     public function definition() {
         global $questionnaire;
+        // We need to get the number of current feedbacksections to allow the rose and radar chart types or not.
+        global $DB;
+        $sid = $questionnaire->survey->id;
+        $sql = 'SELECT COUNT(*) ' .
+           'FROM {questionnaire_fb_sections} ' .
+           'WHERE surveyid = '.$sid;
+        $numfeedbacksections = $DB->count_records_sql($sql, [$sid]);
+
+        // Get all questions that are valid feedback questions.
+        $nbvalidquestions = 0;
+        foreach ($questionnaire->questions as $question) {
+            if ($question->valid_feedback()) {
+                $nbvalidquestions++;
+            }
+        }
 
         $mform =& $this->_form;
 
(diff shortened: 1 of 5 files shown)


classes/feedback_form.php
```diff
@@ -36,6 +36,21 @@ class feedback_form extends \moodleform {
      */
     public function definition() {
         global $questionnaire;
+        // We need to get the number of current feedbacksections to allow the rose and radar chart types or not.
+        global $DB;
+        $sid = $questionnaire->survey->id;
+        $sql = 'SELECT COUNT(*) ' .
```
mchurchward: This should be simplified to a:
$numfeedbacksections = $DB->count_records('questionnaire_fb_sections', ['surveyid' => $sid]);
No need to create an SQL statement.

classes/feedback_form.php
```diff
@@ -44,7 +59,9 @@ public function definition() {
         $feedbackoptions = [];
         $feedbackoptions[0] = get_string('feedbacknone', 'questionnaire');
         $feedbackoptions[1] = get_string('feedbackglobal', 'questionnaire');
-        $feedbackoptions[2] = get_string('feedbacksections', 'questionnaire');
+        if ($nbvalidquestions > 1) {
+            $feedbackoptions[2] = get_string('feedbacksections', 'questionnaire');
+        }
```
mchurchward, line 61: Does $feedbackoptions[2] need to be set to an empty string if the count is one or less?

mchurchward (changes requested):
Can Behat tests be updated/added to verify these?

classes/feedback_form.php
mchurchward: Also, when I looked at the results of an existing feedback, I got the message:
<b>Deprecated</b>:  explode(): Passing null to parameter #2 ($string) of type string is deprecated in <b>/mod/questionnaire/drawchart.php</b> on line <b>107</b><br />

rezeau: I don't understand how this can happen. Anyway, I've added a check to avoid potential error message: if ($label !== null) ...

rezeau:
@mchurchward wrote "Can Behat tests be updated/added to verify these?"
I've checked that the existing Behat test _add_multi_feedback_with_sections.feature_ works OK with this PR. 
Do you prefer me to update that Behat test with my PR new features OR create a new test?

mchurchward:
Hi Joseph. Sorry it has taken so long to get back to this.
I'm having trouble testing it. I have existing questionnaires with feedback options already in the database.
The questionnaire has a normal rate question.
But when I try to access the feedback for editing, I get the message:
_"Feedback options are available if your questionnaire contains the following question types and question settings: Radio buttons; Dropdown box; Yes/No; or Rate (normal or Osgood scale). Those questions must be set as Required, their Question Name field must NOT be empty and the Possible answers choices must contain a value. "_
This error happens on the current MOODLE_401_STABLE branch as well as the one you have provided here.

What am I doing wrong?

rezeau:
@mchurchward I've just tested my PR #510 on my local moodle test site running Moodle 4.3.3+ (Build: 20240215). It works as expected.
You write:: "_The questionnaire has a normal rate question.
But when I try to access the feedback for editing, I get the message:
"Feedback options are available if your questionnaire contains the following question types and question settings: Radio buttons; Dropdown box; Yes/No; or Rate (normal or Osgood scale). Those questions must be set as Required, their Question Name field must NOT be empty and the Possible answers choices must contain a value. ""_
1.- Are you positive that your questionnaire obeys _all the conditions_ mentioned in the Feedback options warning message?
2.- If you try your to add Feedback to your questionnaire _without_ using my PR, are you getting the same message? My PR has nothing to do with those Feedback requirements, i.e. it does not add any extra requirement to the list.

rezeau:
@mchurchward Should I wait for #542 and @lucaboesch 's suggestion :: "_A branch MOODLE_403_STABLE in your repository would be highly appreciated_" before continuing testing this #510 ?

joostelshoff:
@mchurchward What is the current status of this fix?

benjamin-feron:
Hello, is there any chance to see this PR merged ?

rezeau:
@benjamin-feron Probably very little chance.

PR #511, opened 2023-09-28
Questionnaire/Accessibility: Radio buttons & Yes/No missing group label

emanoylov:
Changes to make the Radio buttons and Yes/No buttons better-associated with the question text asked for accessibility.

E.g. if the question is "Select a favourite colour" and the options are Blue/Red/Green the possible responses are read out by the screen reader as:

"Select a favourite colour Blue"
"Select a favourite colour Red"
"Select a favourite colour Green"

The response templates are also modified for consistency.

### DIFF
--- a/classes/question/radio.php
+++ b/classes/question/radio.php
@@ -122,6 +122,9 @@ class radio extends question {
                 }
                 $contents = questionnaire_choice_values($choice->content);
                 $radio->label = $value.format_text($contents->text, FORMAT_HTML, ['noclean' => true]).$contents->image;
+                if (!empty($this->qlegend)) {
+                    $radio->alabel = strip_tags("{$this->qlegend} {$radio->label}");
+                }
             } else {             // Radio button with associated !other text field.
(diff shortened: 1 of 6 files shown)


mchurchward:
I'm not sure about this change. if the question is long and the answers are short, the reading will be very long as well. Is there an accessibility guideline that encourages this?

emanoylov:
@mchurchward This came up from the accessibility testing by our business people and follows the same pattern already implemented for the checkboxes via [#501](https://github.com/PoetOS/moodle-mod_questionnaire/pull/501).

emanoylov:
@mchurchward Here is the accessibility guideline:

The missing group labels fails WCAG 2.1 Success Criterion 1.3.1, [info and relationships] (https://www.w3.org/TR/WCAG21/#info-and-relationships) (level A).

Can you please progress it?

mchurchward (approved):
I'm going to pass this. Thanks.

PR #512, opened 2023-10-02
Fix wrong param declaration in webservice

vuvanhieu143:
Steps to reproduce:
- Go to Server > Web services > Overview
-  Click on "Select a service" Link
- Click the add link
- Create a sample web service
- Click add function
- We will see the warning
image
So this change is to correct the declaration of the web service param and fix the warning message.

### DIFF
--- a/externallib.php
+++ b/externallib.php
@@ -137,7 +137,7 @@ class mod_questionnaire_external extends external_api {
     public static function submit_questionnaire_response_returns() {
         return new external_single_structure(
             [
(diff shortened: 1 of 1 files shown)


mchurchward:
Using the steps above, I don't get that error. Is this specific to a Moodle / PHP version?

vuvanhieu143:
> Using the steps above, I don't get that error. Is this specific to a Moodle / PHP version?
Hi @mchurchward ,
The version we are using is M 4.2
and they are already fixed in the #508 
so I will close this PR.
Sorry about this, I didn't see that change, we are still using the old version(2022092202) so we are going to update the code.

PR #517, opened 2023-10-04
Update sectiontext questions to display on mobile #482

sayoder-aprende:
Also refactored the code a bit for clarity: The question class now has an is_numbered() method (default true) that can be overridden for question types that should not have numbers. Replaced some checks for specific question types with checks for is_numbered().

(Re-creating this pull request to get fresh test runner output)


sayoder-aprende:
For whatever reason I'm still having trouble getting my local PHPDoc to complain, but I hope this change should do it. (Thanks for re-approving the workflow!)

mchurchward (approved):
Thanks for the change. Great to add the new method too.

PR #527, opened 2023-10-20
Allow questionnaire to be gradable LTI advantage activity

liamgm:
In 2013, the FEATURE_GRADE_HAS_GRADES was toggled to False without comment. This makes the questionnaire activity ineligible to send grades when using Moodle as an LTI 1.3 Tool, among other potential effects. Additionally, the grade line item isn't created when a questionnaire is initially added to a course in spite of that initial configuration allowing it to be graded; the line item is, however, added when the activity configuration is re-saved without making changes.

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -42,7 +42,7 @@ function questionnaire_supports($feature) {
         case FEATURE_COMPLETION_HAS_RULES:
             return true;
         case FEATURE_GRADE_HAS_GRADE:
-            return false;
+            return true;
         case FEATURE_GRADE_OUTCOMES:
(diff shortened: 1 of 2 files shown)


mchurchward:
This was changed due to the [tracker CONTRIB-4274 report](https://tracker.moodle.org/browse/CONTRIB-4274).

Can you review there and confirm that this does not cause a regression?

liamgm:
Good find: I will put that on my to-do list and see what happens when following that procedure.

mchurchward:
Closing due to no activity

PR #536, opened 2023-11-29
GHI535 Support for user identity fields in Download Responses

emanoylov:
A new option "Show user identity" in the questionnaire | downloadoptions setting to allows including/excluding the user identity fields in Download Responses.
Additionally if a questionnaire is anonymous the identity fields are excluded in a download even if "Show user identity" is on.

### DIFF
--- a/db/upgrade.php
+++ b/db/upgrade.php
@@ -997,6 +997,11 @@ function xmldb_questionnaire_upgrade($oldversion=0) {
         upgrade_mod_savepoint(true, 2022092200, 'questionnaire');
     }
 
+    if ($oldversion < 2022121600.02) {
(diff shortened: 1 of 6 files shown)


jason-platts:
Thanks. I am not sure on the new version number... @mchurchward do you like having the point number incremented only rather than the whole version update to later date?

mchurchward:
> Thanks. I am not sure on the new version number... @mchurchward do you like having the point number incremented only rather than the whole version update to later date?

Yes. I keep the date part of the version number static for the release branch. So for MOODLE_401_STABLE, we use "20221216xx". And within the branch, add a decimal increase, so "2022121600.02". When I do a new release, I will increment the version completely to something like "2022121601".

mchurchward (approved):

PR #516, opened 2023-10-04
display description on course page MOODLE_401_STABLE

rezeau:

PR #537, opened 2023-12-12
Show description on course page when enabled

srobotta:
Add module decription in course page:


srobotta:
Luca just pointed me to https://github.com/PoetOS/moodle-mod_questionnaire/pull/516 which seems to fix the identical problem.

mchurchward:
Fixed in #516

PR #541, opened 2024-01-18
Circumvent number on string error, resolves #538.

### LINKED ISSUE #538: number on string error
We're seeing various users reporting "attempt to define property number on string" when filling out a form.

Screenshot 2023-12-18 at 15 05 32

It's a numeric question type, so I'm not sure why it's expecting a string (assuming that's what this means) and this has happened since we switched to M4. They are trying to enter a decimal value. Wierdly seems to happen for some users and not others, with no rhyme or reason to this.

lucaboesch:
Referring to the solution given in https://github.com/lucaboesch/moodle-mod_questionnaire/pull/new/issue-538.
Thanks.

### DIFF
--- a/classes/question/numerical.php
+++ b/classes/question/numerical.php
@@ -83,7 +83,7 @@ class numerical extends question {
         // Numeric.
         $questiontags = new \stdClass();
         $precision = $this->precise;
-        $a = '';
+        $a = new \stdClass();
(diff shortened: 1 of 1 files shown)


mchurchward:
This was fixed in #539

lucaboesch:
stdClass is spelled with a lower case s.

PR #547, opened 2024-02-23
Fix for issue sectiontext qtype (label) should NOT support feedback

mchurchward:


mchurchward:
Ah, just found #514. Closing

PR #542, opened 2024-01-19
Prepare for Moodle 4.3 and PHP 8.2.

lucaboesch:
I have a working branch here that fixes #530.

PR #471, opened 2023-03-16
File submission question type

martin-call-learning:
Hi,
Here is an improvement where we added a file question type.
This type allows the user to submit a pdf or an image file to the questionnaire.
FIxes #287 

Thanks for your time reviewing this pull request.

Kind regards.

PR #559, opened 2024-03-25
File submission 403

srobotta:
This pull request combines:
- File upload question type: which is originally from Laurent David
- Behat tests for that feature from Luca Bösch
- Make it ready for Moodle 4.3 from Luca Bösch
- Fixes of mine:
  - there can be more than one upload file question in a questionnary
  - the mandatory field check works as expected


mchurchward:
The pull request should just contain the code it deals with. And it needs to be rebased against the latest MOODLE_401_STABLE. And, what exactly does this address?

srobotta:
Hi Mike,

The code contains the new question type "file upload". We wanted/needed this feature. However, this is not yet part of the official release. The original work has been done by Laurent David, but when we tried out the code, we noticed a lot of issues. That's why I did changes based on his work. Apart from this, Luca maily wrote Behat tests to cover the new feature.
Before doing the pull request, I did merge the MOODLE_401_STABLE branch, so it should be up to date.

Best regards, Stephan

mchurchward:
Okay. Then you need to change this PR. PR's should only be for the feature they deal with. Otherwise it is too complicated to deal with. The commits that were for general unrelated fixes (such as Luca's CI and 4.3 fixes) should not be part of this.
Luca's changes are in a [different PR](https://github.com/PoetOS/moodle-mod_questionnaire/pull/542).

srobotta:
Hi Mike,

The problem that I see here is that the original pull request #471 which we used and discovered that it was not correctly working. I myself have no knowledge of how I could extend this original pull request, therefore I took the changes, applied my fixes and created a new pull request. Sorry for the hazzle.
Lucas #542 deals with the 4.3 compatibility, the changes from him in this pull request acutally add Behat tests for the file type question.
The whole package can be tested at once without the need of having several pull request to test which patially would even fail because they depend on each other.
Therefore my question to you, how can we proceed here that both, you and me do not have much effort in applying the changes?

Best regards, Stephan

mchurchward (changes requested):
Sorry for the long delay. This needs to be split up into multiple pull requests. Note that one of the PR's you reference from Luca has already been merged. I would suggest that you do a new PR off of the current MOODLE_401_STABLE branch that deals only with the File submission feature. Any other features/fixes need to likewise be in new PR's.

mchurchward:
I may have broken your rebase. Can you do it one more time?

srobotta:
I have finished my work, and will create a new pull request.

PR #562, opened 2024-03-28
Delete all incompletes when deleting all responses

Syxton:


mchurchward (approved):
Thank you for this fix.

PR #563, opened 2024-04-04
Circumvent number on string error, resolves #538.

### LINKED ISSUE #538: number on string error

lucaboesch:


mchurchward (approved):
(looks for the palm to face smiley)
Thanks for fixing this.

PR #581, opened 2024-07-22
Adapt mobile code to ionic 7

dpalou:
In this PR I'm removing the ionic3 code (the ionic5 app was released on 2021 so it should be safe to remove it).

I also created a new latest template to use the new Ionic syntax for inputs. The old syntax still works in the 4.4 version of the app (ionic 7), but it will stop working in the 4.5 version (ionic 8) so some inputs could be wrongly aligned without these changes. To make it easier to review the code, in the first commit I duplicate the template (no changes done in the HTML) and then I apply the changes in a new commit.

Finally, I also removed the usage of core-course-module-description component for newer app versions, it is deprecated and will be removed in a future version of the app. It should have been removed already, but we kept it in the app because some plugins still use it.


mchurchward (commented):
Thanks for this. So, just confirming. These changes will work on all current Moodle 4 releases, correct?

dpalou:
Hi @mchurchward ,

yes, almost all of the changes are in Mustache templates that are only used for the app, they don't have any effect in Moodle LMS. The only change done in PHP code is a variable value, so it's safe.

The plugin will stop working in apps still using Ionic 3 (because I removed the ionic3 templates), but we upgraded to Ionic 5 more than 3 years ago so it should be safe to remove it.

mchurchward (approved):

PR #582, opened 2024-07-25
Add 4.4 and main to matrix

aspark21:


aspark21:
tests fail - https://github.com/aspark21/moodle-mod_questionnaire/actions/runs/10101874369

that's the point

mchurchward (changes requested):
Thank you for this. Can you rebase with the latest MOODLE_401_STABLE to re-enable the CI tests?
I'm not sure we want 'main' yet.

mchurchward:
I have [added 4.4 to the matrix](https://github.com/PoetOS/moodle-mod_questionnaire/commit/33096037711206df48e8ce86536414cca99f8453). I don't want "main" in there as it is not released yet.

aspark21:
Yeah main will never be released but the point is that it gives you early visibility into things that need to change for the next version. If those tests fail you can always ignore them for a while.

At least we have found it useful with our own plugins

PR #589, opened 2024-08-15
Questionnaire: make Summary page display more intuitive

toanlamt:
E.g. In this questionnaire, they are 151 questionnaire responses, but 198 question responses. As only complete questionnaire responses are analyzed (the data is exported and cleaned), showing more question responses than questionnaire responses in the summary is confusing.

E.g. here's a view of a recent case:
image

If we could make the view selectable, this would clean up the summary and make a bit more sense.

Drop-down: "All responses" [198 in this case]; "Full submissions" [151 in this case]; "Responses not submitted" [47],
Consistent language to delineate between question-specific responses and full/complete questionnaire submissions.
Correct the 'total' and percentages on the visual graphs depending on the view.
E.g. here's a view after improvement
image


mchurchward:
Can you rebase with the latest MOODLE_401_STABLE? I just fixed the CI.

mchurchward (changes requested):
Your PR still includes commits from other changes that this one. The PR needs to just include the one(s) for this PR.
You need to pull the latest MOODLE_401_STABLE locally, then rebase your branch on top of that.

tailetan:
rebased with MOODLE_401_STABLE, @mchurchward

mchurchward (dismissed):

classes/responsetype/responsetype.php
```diff
@@ -198,7 +198,7 @@ public function get_results_tags($weights, $participants, $respondents, $showtot
             if ($showtotals) {
                 $pagetags->total = new \stdClass();
                 if ($respondents > 0) {
-                    $percent = round((float)$respondents / (float)$participants * 100.0);
+                    $percent = 100;
```
mchurchward: Why this change? This changes a calculation to a hard coded value.

questionnaire.class.php
```diff
@@ -855,15 +857,20 @@ public function get_responses($userid=false, $groupid=0) {
             $sql = 'SELECT r.* ' .
                 'FROM {questionnaire_response} r ' .
                 $groupsql .
-                'WHERE r.questionnaireid = :questionnaireid AND r.complete = :status' . $groupcnd;
+                'WHERE r.questionnaireid = :questionnaireid' . $groupcnd;
```
mchurchward, line 856: Why did you remove the 'complete' check?

mchurchward (commented):
Left a couple of questions.

classes/responsetype/responsetype.php
toanlamt: Hi @mchurchward 
Sorry for this change. During the development of this issue (previously handled by our former developer), it seems that updating the "All participants" text to a dropdown list with the content "Full submissions" - Default value, "All responses", "Responses not submitted" led us to assume that the "Total responses to question" row should always be the sum of the response number/percent above. For example:

Response: Yes - 1 -> 50%
Response: No - 1 -> 50%
Total responses to question row: (2) -> 100%

After thoroughly investigating the logic of this row, we now understand that it should not be hard-coded to 100 as it was before. We will update this code to align with the current logic.

questionnaire.class.php
toanlamt: Due to the new requirement of separating the 3 response viewing modes, we have been updating this SQL query and adding some logic to check the status correctly for each view. 

image

mchurchward:
I have replaces this with a new https://github.com/PoetOS/moodle-mod_questionnaire/pull/621  and rebased it with MOODLE_404_STABLE.

mchurchward:
Rebased with MOODLE_404_STABLE in this [PR](https://github.com/PoetOS/moodle-mod_questionnaire/pull/621).
Closing this one.

PR #590, opened 2024-08-15
Questionnaire: option to auto-delete responses after X time

toanlamt:
For the purpose of having a GDPR-friendly feature (disabled by default) for the Questionnaire, we are suggesting this pull request. The approach and purpose are potentially similar to the ticket for Quiz activity, [MDL-69854](https://tracker.moodle.org/browse/MDL-69854), which you can refer to for details.
For easy visualization, here’s what we are proposing:
At the bottom of the current list for the Questionnaire setting, we are adding a new option called "Manage old responses." With this option, the system will automatically delete student responses after a specified amount of time has passed (including both partial and submitted attempts).

  - **Value**: Never remove; 1-36 months
  - **Default value**: Never remove

Depending on the value selected by the user, within the corresponding time frame, the system will automatically remove discussions if they haven't had any new replies for a certain length of time.
This helps free up memory by removing outdated data, potentially improving system performance slightly, while also reducing the amount of stored data to mitigate GDPR risks.


mchurchward:
I'm not sure I agree with this change. If I'm understanding correctly, it will delete responses automatically in the background if they are older than a specified time? What is the use case for this?

mchurchward:
This issue needs to describe what it is for, and how it provides the solution. Use cases / user stories would be helpful.
Does this type of functionality exist in any Moodle core activities?

mchurchward:
I will accept this if a global questionnaire setting is added to enable/disable this new function. It should be defaulted to disabled, requiring any site to turn it on before it takes affect.
Also, there are new conflicts that need to be fixed.

tailetan:
I will update and get back to you soon. Thanks

mchurchward:
And one more requirement, sorry. Any specific questionnaire should be able to prevent it's responses from being deleted automatically. That is, if the automatic deletion is enabled for questionnaire globally, then any questionnaire can have a setting that prevents this. And, existing questionnaires should have that disabled by default to protect them.

tailetan:
I have just updated the commit based on the @mchurchward's requirement, please review again. Thanks

locallib.php
```diff
@@ -949,3 +949,63 @@ function questionnaire_get_standard_page_items($id = null, $a = null) {
...
+ * @throws dml_exception
+ */
+function questionnaire_delete_old_responses() {
+    global $DB;
+    $currenttime = time();
+
+    $sql = "SELECT qr.id
+              FROM {questionnaire} q
+              JOIN {questionnaire_response} qr ON qr.questionnaireid = q.id AND qr.complete = 'y'
+             WHERE q.removeafter <> 0 AND (q.removeafter < :currettime - qr.submitted)";
```
mchurchward: :currettime should be :currenttime

locallib.php
```diff
@@ -949,3 +949,63 @@ function questionnaire_get_standard_page_items($id = null, $a = null) {
...
+            }
+            // Delete all of the response data for a response.
+            $responsetables = [
+                    'questionnaire_response_bool', 'questionnaire_response_date', 'questionnaire_resp_multiple',
+                    'questionnaire_response_other', 'questionnaire_response_rank', 'questionnaire_resp_single',
+                    'questionnaire_response_text'];
+            list ($sqlparam, $params) = $DB->get_in_or_equal($oldresponsesid, SQL_PARAMS_QM);
+            foreach ($responsetables as $tablename) {
+                $sql = "DELETE FROM {{$tablename}} WHERE response_id $sqlparam";
+                $DB->execute($sql, $params);
```
mchurchward: Use a Moodle $DB 'delete_records' API function instead of execute.

mchurchward (changes requested):
See inline comments.

tailetan:
Ready for review @mchurchward

mchurchward:
> Ready for review @mchurchward

Thanks. In the future, don't squash a review commit fix. It is very difficult to determine what you changed since the review when that is done. Instead, leave them as new commits in the PR, and they can be squashed with the PR.

mod_form.php
```diff
@@ -223,4 +237,18 @@ public function completion_rule_enabled($data) {
         return !empty($data['completionsubmit']);
     }
 
+    /**
+     * Create options for remove old responses in the questionare.
+     *
+     * @return array
+     */
+    public function create_remove_options() {
```
mchurchward, line 245: This function doesn't appear to be used anywhere. It may be a duplicate of locallib::questionnaire_create_remove_options()? If so, it may make sense to use this one rather than the one in locallib?

lang/en/questionnaire.php
```diff
@@ -570,6 +576,10 @@
 $string['resume_link'] = 'mod/questionnaire/mod#Save/Resume_answers';
 $string['resumesurvey'] = 'Resume questionnaire';
 $string['return'] = 'Return';
+$string['removeoldresponsesdefault'] = 'Never remove';
+$string['removeoldresponses'] = 'Manage old responses';
+$string['removeoldresponsesafter'] = 'Manage old responses after';
```
mchurchward: This should use "Delete" or "Remove" instead of "Manage" for clarity.

settings.php
```diff
@@ -52,4 +53,19 @@
...
     $settings->add(new admin_setting_configcheckbox('questionnaire/allowemailreporting',
         get_string('configemailreporting', 'questionnaire'), get_string('configemailreportinglong', 'questionnaire'), 0));
+
+    // Manage old responses after. The default value is 24 months.
+    $options = questionnaire_create_remove_options();
+    $settings->add(new admin_setting_configselect('questionnaire/removeoldresponses',
+            get_string('removeoldresponsesafter', 'questionnaire'),
+            get_string('configremoveoldresponses', 'questionnaire'), 0, $options));
+
+    $options = [
```
mchurchward, line 58: Ideally, this setting should appear before the previous setting ("questionnaire/removeoldresponses"), since it directly impacts that one.

backup/moodle2/backup_questionnaire_stepslib.php
```diff
@@ -39,7 +39,7 @@ protected function define_structure() {
         $questionnaire = new backup_nested_element('questionnaire', array('id'), array(
             'course', 'name', 'intro', 'introformat', 'qtype',
             'respondenttype', 'resp_eligible', 'resp_view', 'notifications', 'opendate',
-            'closedate', 'resume', 'navigate', 'grade', 'sid', 'timemodified', 'completionsubmit', 'autonum'));
+            'closedate', 'resume', 'navigate', 'grade', 'sid', 'timemodified', 'completionsubmit', 'autonum', 'removeafter'));
```
mchurchward, line 42: Have you tested a restore of an older questionnaire backup that doesn't have the removeafter setting to ensure there is not an issue?

mchurchward (changes requested):
Please review and address inline comments as needed.

backup/moodle2/backup_questionnaire_stepslib.php
tailetan: I have just tested and it works as expected

tailetan:
@mchurchward this is now ready for review again.

mchurchward:
Sorry. There is a conflict I can't see for some reason. Can you check that your PR branch is up to date with the current MOODLE_404_STABLE branch?

tailetan:
@mchurchward Just rebased. Please review it again. Thanks

mchurchward:
This was resolved with [PR634](https://github.com/PoetOS/moodle-mod_questionnaire/pull/634)

Issue #594, opened 2024-08-22
JavaScript error formChild is null on completion page


leonstr:
### Steps to reproduce issue

1. Create a course, enrol a teacher and student.
2. As the teacher:
3. Add a Questionnaire to the course.
4. Click _Add questions_ and add a question.
5. As the student click on the Questionnaire on the course page, then click _Answer the questions..._.
6. As the student answer the question and click _Submit questionnaire_.

### Expected results

No errors in browser console.

### Actual results

Error on browser console: _Uncaught TypeError: formChild is null_.

### Environment

This error occurs in Moodle 4.0, 4.1 and 4.2 with mod_questionnaire 2022092202. It occurs with Moodle 4.3.3 and later, and Moodle 4.4 with PR #584 (needed to get the JavaScript to run at all).

leonstr:
PR #595 has a proposed fix for this.

mchurchward:
Fixed in #595

PR #595, opened 2024-08-22
594 Fix formChild is null

leonstr:
When a student completed a questionnaire then shown the "Thank you for completing this Questionnaire" page there was a JavaScript error:

  Uncaught TypeError: formChild is null

This was because the JavaScript for the questions form was being inadvertently loaded for the Thank you page where this form is not present.

### DIFF
--- a/classes/output/renderer.php
+++ b/classes/output/renderer.php
@@ -147,6 +147,7 @@ class renderer extends \plugin_renderer_base {
         foreach ($hiddeninputs as $name => $value) {
             $output .= \html_writer::empty_tag('input', ['type' => 'hidden', 'name' => $name, 'value' => $value]) . "\n";
         }
+        $this->page->requires->js_init_call('M.mod_questionnaire.init_attempt_form', null, false, questionnaire_get_js_module());
         return $output;
     }
 
--- a/questionnaire.class.php
+++ b/questionnaire.class.php
@@ -270,10 +270,6 @@ class questionnaire {
 
         $PAGE->set_title(format_string($this->name));
         $PAGE->set_heading(format_string($this->course->fullname));
-
-        // Initialise the JavaScript.
-        $PAGE->requires->js_init_call('M.mod_questionnaire.init_attempt_form', null, false, questionnaire_get_js_module());
-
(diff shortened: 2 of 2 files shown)


mchurchward (changes requested):
This will need to rebased with the latest MOODLE_401_STABLE so that only this PR's commit is present.

leonstr:
Apologies, rebased, hopefully this is correct now.

Edit: No it isn't. More apologies. I'll see if I can sort this out.

leonstr:
Maybe it is correct after all, please let me know if there are still issues with this.

mchurchward:
There was a fix added to the [javascript here](https://github.com/PoetOS/moodle-mod_questionnaire/commit/852486f30b967d5d61847dfaff9ec8cd23f4c11f). This may already fix the problem you describe?

leonstr:
> There was a fix added to the [javascript here](https://github.com/PoetOS/moodle-mod_questionnaire/commit/852486f30b967d5d61847dfaff9ec8cd23f4c11f). This may already fix the problem you describe?

No, in fact this issue only occurs once that fix has been applied. Fix 852486f allows the form checker to be run with Moodle 4.4 (instead of getting a console error), but this means a long-time bug re-emerges due to the form checker being incorrectly loaded for the "Thank you" page. If in doubt, install the latest MOODLE_401_STABLE of mod_questoinnaire and follow the steps in #594.

mchurchward:
I don't get that error using the latest MOODLE_401_STABLE.

leonstr:
> I don't get that error using the latest MOODLE_401_STABLE.

On the Thank You page? That's a mystery since there's no DOM element with the ID specified by the form checker code.

I've just retested, on Moodle 4.4.2+ (Build: 20240823) with mod_questionnaire 5c0dbe8 and still see the error if I follow the steps from #594 (slightly different error because it's a different browser):

typeerror

mchurchward:
Ah. Okay. I didn't realize you meant in the console. Yes, I do see that.

mchurchward (approved):
Thanks for the clarification.

PR #604, opened 2024-09-10
Add free text (Other) to rate/scale.

toanlamt:
**AC1**: When option 'Other' is added in Scale/Rating question type, the 'Other' option should display with textbox next to it in the question view

**AC2**: When user answer the Scale/Rating question type question with 'Other' responses, it should display the other response as one of the options within the question in:
- Your Response view
- Your Response/Individual responses view
- Your Response/ responses view

**AC3**: When there are multiple responses for that Scale/Rating question: 
In Individual Response view, it will display each individual response per page
In All Response view, it will display each response separately
in the summary view, keep all the input of ‘other’ options as the ‘Other’ category. Therefore it will show as 1 option. 

**AC4**: The other option in Scale/Rating question type should display in Summary view also with the calculation logic: if a user inputs ‘Other’ and ranks it, it should be the rank average of the ‘Other’ category. The calculation for the point and percentage follow the same as existing behaviour (normal option)
Note:
 1. Average rank for each option: sum point of each time rating / total times of rating
 2. Responses point: the percentage of the times option is chosen.

### DIFF
--- a/classes/question/rate.php
+++ b/classes/question/rate.php
@@ -336,8 +336,22 @@ class rate extends question {
                 if ($this->osgood_rate_scale()) {
                     list($content, $contentright) = array_merge(preg_split('/[|]/', $content), array(' '));
                 }
-                $cols[] = ['colstyle' => 'text-align: '.$textalign.';',
-                           'coltext' => format_text($content, FORMAT_HTML, ['noclean' => true]).'&nbsp;'];
+                if ($choice->is_other_choice()) {
+                    $othertext = $choice->other_choice_display();
+                    $oname = $cid . '_qother';
+                    $oid = $cid . '-other';
+                    $odata = isset($response->answers[$this->id][$cid]) ? $response->answers[$this->id][$cid]->value : '';
+                    if (isset($odata)) {
+                        $ovalue = stripslashes($odata);
+                    }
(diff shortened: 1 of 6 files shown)


mchurchward:
I like this idea. But I wonder if the all responses summary page should list all of the "Other" choices individually like is done for check boxes:
Screenshot 2025-02-20 at 1 50 59 PM

Currently, this change lists them as a comma separated list on one line:
Screenshot 2025-02-20 at 1 51 08 PM

This could quickly get long and unreadable.

mchurchward (changes requested):
See my [comment](https://github.com/PoetOS/moodle-mod_questionnaire/pull/604#issuecomment-2672399270)

toanlamt:
Hi @mchurchward,

Thank you for your feedback. I think listing all of the 'Other' choices individually on the response summary page makes sense to me.

I have updated my code and created a separate [commit](https://github.com/PoetOS/moodle-mod_questionnaire/compare/MOODLE_404_STABLE...NashTechOpenUniversity:moodle-mod_questionnaire:wip313483) to fix this issue. Here is a screenshot of the response summary page after the adjustment. Could you please review it again?

image

mchurchward (approved):

PR #611, opened 2024-10-09
Suppres huge size of icons in Manage questions in Moodle 4.5

### LINKED ISSUE #612: Serious display issue in Moodle 4.5
Moodle 4.5 - Questionnaire 4.1.1 (Build - 2024082900) 2022121601 - using Moodle default theme Boost
When adding questions to a questionnaire, the Manage Questions section displays huge icons, which makes the creating interface almost un-usable. Please provide a fix.

tats-u:
I upgraded my Moodle to 4.5 and found the icons in Manage questions page are super huge. 
It's due to the icons were changed from tiny raster ones to huge SVG ones in 4.5.
We now have to specify their size in CSS.

Fixes #612

### DIFF
--- a/styles.css
+++ b/styles.css
@@ -11,6 +11,10 @@
     color: gray;
 }
 
+#page-mod-questionnaire-questions .qcontainer .qnums ~ input[type=image] {
+    max-height: 1em;
+}
+
 #page-mod-questionnaire-questions .qcontainer .fstatic {
     width: 97%;
     margin-right: 1em;


rezeau:
Tested this PR #611 and I confirm that it works as expected.

styles.css
```diff
@@ -11,6 +11,10 @@
     color: gray;
 }
 
+#page-mod-questionnaire-questions .qcontainer .qnums ~ input[type=image] {
+    max-height: 1em;
```
tats-u, line 15: `height: 1em; width: auto;` might be better

I adopted max-height to keep visual compatibility with prior versions but I doubt a little that it's the best.

david-jpb: Hello

I also have this same issue

I've tried

#page-mod-questionnaire-questions .qcontainer .qnums ~ input[type=image] {
    max-height: 1em;
}

and 

#page-mod-questionnaire-questions .qcontainer .qnums ~ input[type=image] {
    height: 1em; 
    width: auto;
}

And it still occurs

Just to make sure, the idea is to edit the styles.css in the Questionnaire folder, correct?

I'm on questionnaire 4.1.1 (Build - 2024082900) and using Moodle 4.5

Thank you

tats-u: Have you tried injecting it on Moodle as an additional HTML if you're using the stable version instead of my branch here?

david-jpb: Adding that on additionalhtmlhead

Seems to have done the trick

Yeah, I still get a bit lost around the different versions one might encounter when using GIT, so that might not have helped

Thanks a lot for the help, I can finally do my testing without destroying my mouse scroll wheel 💪

mchurchward (approved):

Issue #626, opened 2025-03-03
[v4.1.1] [Moodle 4.4.6] [PHP 8.1] course/defaultcompletion.php - Form validation error when checking and saving with custom completion rule "submit"


danowar2k:
Moodle 4.4 requires custom completion rules to have a suffix "_$modname".

When Moodle checks if automatic completion is enabled here: https://github.com/moodle/moodle/blob/b132ff1ed83dbe243909b8b121304ab20c0f069c/completion/classes/form/form_trait.php#L403

mod_questionnaire checks the given $data object only for the rule name without the suffix:
https://github.com/PoetOS/moodle-mod_questionnaire/blob/666310decff7c4bc8cc3278dbe59a480f4988640/mod_form.php#L222-L224

while earlier it did create the form elements with the suffix: https://github.com/PoetOS/moodle-mod_questionnaire/blob/666310decff7c4bc8cc3278dbe59a480f4988640/mod_form.php#L211-L214

This leads to the bug that the custom completion rule can't be activated and therefore not be used (at least as a default)

minhduchoang195:
I have already created a pull request https://github.com/PoetOS/moodle-mod_questionnaire/pull/630 to solve this problem.

PR #630, opened 2025-04-07
#626 Fixed issue where custom completion rules cannot be saved

minhduchoang195:
Due to the missing suffix name `completionsubmit` in the validation function `completion_rule_enabled`, the custom completion rules cannot be saved: 
The custom completion rule can't be activated because of missing suffix

Solution: The name of this  custom completion rule should be followed by a similar suffix everywhere.


mchurchward:
Thanks. Can you add a behat test that shows the failure and the fix please?

PR #631, opened 2025-04-10
Add group filter dropdown to List of responses

izendegi:
We've detected that when viewing the _View all responses > Summary_ on a questionnaire with group mode enabled you can filter the group. On the other hand, the _List of responses_ tab doesn't have that option, but the group filter is applied if the user has selected a group on the other tab.

So, this commit adds the group filter to the List of responses tab so that it can be changed directly on that tab.

### DIFF
--- a/report.php
+++ b/report.php
@@ -727,6 +727,13 @@ switch ($action) {
             }
         }
 
+        // Add group filter dropdown.
+        if ($groupmode > 0) {
+            $groupselect = groups_print_activity_menu($cm, $url->out(), true);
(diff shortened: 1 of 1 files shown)


mchurchward (approved):
I've created a new branch and [pull request](https://github.com/PoetOS/moodle-mod_questionnaire/pull/674) with your commit. that avoids the conflicts with the newer main branch. I also added a commit to fix the blank page that occurs if you go to the by response page without any responses present.

mchurchward:
Moved to [pull request 674](https://github.com/PoetOS/moodle-mod_questionnaire/pull/674).

Issue #619, opened 2024-12-24
Questions can have negative position numbers


HirotoKagotani:
I had the same situation as a discussion topic https://moodle.org/mod/forum/discuss.php?d=390636 and I figured out how to give negative position numbers to questions.

I am using Questionnaire version 4.1.1 on Moodle 4.1.

Reproduction steps:
1. Add a Questionnaire Q1 with one or more question(s).
2. Add another Questionnaire Q2 with:
   - a question.
   - a page break.
   - any number of question(s).
3. Move the 1st question of Q2 to the next position of the page break.
4. The page break is deleted by the code in class/questions_form.php.
5. You see the position number of the 1st question of **Q1** is **0**. (Also the position numbers of ALL questions of ALL questionnaires are decreased by one)
6. Add a page break to Q2.
7. Move the page break to the 2nd position.
8. Move the 1st question of Q2 to the next position of the page break again.
9. You see the  position number of the 1st question of **Q1** is **-1**.
10. and so on...

Also we need a method to safely reset the position numbers for all questionnaires (in a Moodle) from 1.

simondipad:
I have this issue, and have discovered that the effect of these negative numbers to the user is that **the save/resume action drops the user back into the questionnaire several pages ahead of where they left off, even if there are previous pages with required responses, these are skipped over**. There is a reset method but the problem recurs in the same questionnaire instance after a while. Tested this on two seperate moodle installations with the same results. This is a major problem from a UX point of view.

PR #639, opened 2025-07-07
Negative position setting fix 

Bartman119:
Fix negative position setting for edited questionnaire and remove position updating for other questionnaires


Bartman119:
Fix regarding #619

mchurchward:
Thanks for this. Can you also provide a Behat test that shows the error before the fix, and no error after the fix?

Bartman119:
Sure, I will provide appropriate behat tests at the earliest convenience

> Thanks for this. Can you also provide a Behat test that shows the error before the fix, and no error after the fix?

HirotoKagotani:
How about using a placeholder just in case?

    if ($records = $DB->get_records_select('questionnaire_question', 'surveyid = :sid', ['sid' => $sid], 'position ASC')) {

PR #650, opened 2025-08-06
Fix numerical response value

DLM-unipd:
Removed the [is_numeric()](https://www.php.net/manual/en/function.is-numeric.php) check since the preg_replace() function already handles the extraction of valid numeric values. This fix properly processes inputs with comma decimal separators and maintains the intended behavior while simplifying the code.

The original code did not handle non-standard numeric values correctly, e.g. using comma as decimal separator (42,6).


mchurchward:
Can you provide a test case that fails prior to this change and passes after please?

PR #651, opened 2025-08-18
Questionnaire: Showing warnings when starting Answer the questions

emanoylov:
OU Ref: 900306

Fix for intermittent warnings when starting "Answer the questions" as follows:

Warning: Undefined array key 1 in /var/www/html/moodle/mod/questionnaire/questionnaire.class.php on line 889

Warning: foreach() argument must be of type array|object, null given in /var/www/html/moodle/mod/questionnaire/questionnaire.class.php on line 889

Warning: Undefined array key 1 in /var/www/html/moodle/mod/questionnaire/questionnaire.class.php on line 1329

Warning: foreach() argument must be of type array|object, null given in /var/www/html/moodle/mod/questionnaire/questionnaire.class.php on line 1329

### DIFF
--- a/questionnaire.class.php
+++ b/questionnaire.class.php
@@ -885,7 +885,7 @@ class questionnaire {
                     return true;
                 }
             }
-        } else {
+        } else if (key_exists($section, $this->questionsbysec)) {
             foreach ($this->questionsbysec[$section] as $questionid) {
                 if ($this->questions[$questionid]->required()) {
                     return true;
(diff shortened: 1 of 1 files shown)


mchurchward:
Have you determined what situation causes this? It would be good to know why that error has occurred.

emanoylov:
We got these warnings when you go to a questionnaire and start answering the questions.
They happened intermittently so the fix is just preventative code which is also used in other places within the plugin.

mchurchward (approved):
It would be best if we knew what caused this, but for now this fix seems okay.

PR #661, opened 2025-09-24
Is siteadmin

### LINKED ISSUE #623: Unknown function isadmin
`isadmin` is called here: https://github.com/PoetOS/moodle-mod_questionnaire/blob/36f22d41a3bcf990e3633bcc4ea381d901ec5f2c/locallib.php#L386
Since it's in a if block that I don't think could ever be true, perhaps we don't need that block.

Alternatively use `is_siteadmin()`

sharpchi:
Fixes #623

### DIFF
--- a/locallib.php
+++ b/locallib.php
@@ -383,7 +383,7 @@ function questionnaire_get_survey_list($courseid=0, $type='') {
     global $DB;
 
     if ($courseid == 0) {
-        if (isadmin()) {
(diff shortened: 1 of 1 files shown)


mchurchward (commented):
If the problem is that the function "isadmin()" doesn't exist, this must throw an error when called. Have you found the situation where this throws an error?

Issue #624, opened 2025-02-20
Variable name changed, but use of old name remained


sharpchi:
It appears that `$data` was the old variable name for `$response` in the function's parameters:
https://github.com/PoetOS/moodle-mod_questionnaire/blob/fc6defdb6f6e7fb429419c0ef2bfeacaf91c42bc/classes/question/rate.php#L213

cf commit:
https://github.com/PoetOS/moodle-mod_questionnaire/commit/5a213a2ae74c1d9b7cd103e4bbfbec3e90b170ee#diff-eef56cf13108196b9f96ef83521979965b80b9ff7a0c41f3ab40be86ac1b5dc6R118

PR #662, opened 2025-09-24
if block with data is not used

sharpchi:
Had a good look, and it appears this code doesn't do anything - doesn't throw any errors either.


mchurchward:
I have changed the base to the latest default branch. Errors will need to be fixed.
BUT - are you sure this is needed? Are you sure this case cannot happen? Seems like an unnecessary and risky change.

lucaboesch:
Are you sure, @sharpchi that the solution isn't to simply change this to $response, which if I understand correctly is what #624 alludes to?

PR #669, opened 2025-11-04
Activities Course overview integration, resolves #668

### LINKED ISSUE #668: Assure "Activities" course navigation item is displaying something reasonable
Image

Image

lucaboesch:


mchurchward:
There are CI issues and conflicts. Can you fix these please?

mchurchward:
Good to go?

mchurchward:
Is there some Moodle Dev documentation that shows the need for these changes? I am unfamiliar with this.

lucaboesch:
Yeah, absolutely. https://moodledev.io/docs/5.0/devupdate#activity-overview-page-integration and https://moodledev.io/docs/5.0/apis/plugintypes/mod/courseoverview and yes it is good to go.

classes/courseformat/overview.php
```diff
@@ -0,0 +1,172 @@
...
+    public function __construct(
+        cm_info $cm,
+        /** @var \core\output\renderer_helper $rendererhelper the renderer helper */
+        protected readonly \core\output\renderer_helper $rendererhelper,
+    ) {
+        parent::__construct($cm);
+        $this->manager = manager::create_from_coursemodule($cm);
+    }
+
+    #[\Override]
```
mchurchward, line 57: What is the purpose of #[\Override]?

index.php
```diff
@@ -22,204 +22,11 @@
...
-
-// Configure table for displaying the list of instances.
-$headings = [get_string('name')];
-$align = ['left'];
-
-if ($showclosingheader) {
-    array_push($headings, get_string('questionnairecloses', 'questionnaire'));
-    array_push($align, 'left');
-}
+// phpcs:disable moodle.Files.MoodleInternal.MoodleInternalGlobalState
```
mchurchward, line 25: Why are these here?

mchurchward (commented):
Very good. I just had a couple of clarification questions.

classes/courseformat/overview.php
lucaboesch: When a method has a PHPDoc block just composed of #[\Override] this means it needs no own PHPDoc block because it is overriding it from an abstract class and the PHPDoc block can be viewed there.

index.php
lucaboesch: If we leave those away, Codechecker says it needs 
`defined('MOODLE_INTERNAL') || die();`
Then again, if we put it in, Codechecker says Internal check is superfluous.

So this disabling of the check is the only way to make Codechecker go green.

mchurchward:
There is a new Behat failure.

mchurchward:
> There is a new Behat failure.

I reran it and it resolved itself.

mchurchward (approved):

